PR #22, opened 2021-04-09
MirageOS support (Unix only)

dinosaure:
This is a first shot to support MirageOS with [paf](https://github.com/dinosaure/paf-le-chien). I would like to explain the difficulty for MirageOS to support this framework, our vision about the abstraction and an explanation about what I did on this PR.

In the spirit of MirageOS, we need to be aware that we don't have:
- a file-system
- a concrete socket
- anything related to a _syscall_

This means above all that our projects must be thought of as interchangeable software bricks depending on the target (from the CPU architecture, the available devices, to the application desired by the user). This is why all our MirageOS projects often work ... everywhere!

However, to make them work specifically for your case, for your application, you often need a little bit of glue. This is done by applying a _functor_, specializing a function, choosing an implementation at the link stage. But a MirageOS project will never go beyond the barrier between what is the implementation of a format, a protocol, the core, etc. and a real and usable example. And if we really need to cross that barrier, it would be by an explicitly separate derivation of the core (confers `*-lwt-unix`, `*-async`, etc.).

Anyway, the point is that the desire to have a _first-entry point_ within `dream` that can run an HTTP server is beyond this barrier. So to speak, the project as it stands is not, by design, compatible with MirageOS. Since the end user would have to use the `Dream` module, it would inevitably come with `Dream.run`. Even if the latter wanted to use MirageOS support (which is a functor), Dream.run would continue to exist in the final object - and would not be able to compile for targets like Solo5 or Xen.

Second, Dream uses dependencies that transitively bring `unix` and `lwt.unix`. Or it uses dependencies that bring C dependencies that may be incompatible (like `cryptokit`). Indeed, in the spirit of MirageOS, we try to remove as much C as possible. This work is necessary because we do not integrate a standard C library! However, MirageOS does allow compiling projects for Unix. In this case (and this is the example given here), Dream works.

Finally, this is mainly about HTTP. Considering everything I just said, it can be difficult to solve this gap between a TCP/IP stack given afterwards and TLS + ALPN while remaining compatible with MirageOS. This is where mimic comes in. We can consider this socket as the only one that is concrete (no _functor_ needed) and compatible with MirageOS. In fact, the resolution of what `Mimic.flow` represents is done dynamically - you can see a tutorial about it [here](https://mirage.github.io/ocaml-git/mimic/index.html).

Thanks to Mimic, we can start to process the information and especially ALPN with `ocaml-tls` (which is compatible with MirageOS). This is what I started doing for [`paf`](https://github.com/dinosaure/paf-le-chien/pull/20) which is just an extension of Mimic to the special case of HTTP. The latter supports HTTP/1.1 with `http/af` and H2 with `h2`.

At this point we are still compatible with MirageOS. This means that we are compatible with all contexts _a priori_. Indeed, using the _functor_ with `tcpip.stack-socket` derives [`serve`](https://github.com/aantron/dream/compare/master...dinosaure:mirage?expand=1#diff-99cba5533ea54eb494c60de86fe54539e9a458231fd90d68080dd02c3390a8d0R2044) to be usable in a simple executable - again, in our way of doing things, we won't go beyond that barrier. A bit more abstract, [`dream.mirage`](https://github.com/aantron/dream/compare/master...dinosaure:mirage?expand=1#diff-56a604a5241790aa4d845f9055ef47a8641756f3cbbf534a6894f1177d8ea6c5) expects only 4 system specific things plus a `Dream.app` and the `Dream.handler`:
- the accept function
- the close function
- the injection of the _flow_ into a Mimic.flow (for this you have to read the tutorial)
- the projection of some value from the flow (ALPN, peer)

So it is possible to derive a `unix` support directly from these 4 informations and directly use a `Lwt_unix.file_descr` if you want. Again, in the spirit of MirageOS, we don't want to go beyond this barrier.

So we have something that works for MirageOS and only for `unix`. This is a first step :) . You can try MirageOS by going to `example/x-mirage` and:
```
$ mirage configure -t unix
$ make depends
$ mirage build
$ ./main.native
```

From the discussion with @aantron. Theoretically Dream can be compatible with MirageOS and we can imagine a complete system with it. This PR is the first step. However, there are still a few steps to go before we have a DreamOS 🙂 !

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -10,3 +10,7 @@
 [submodule "src/vendor/h2"]
 	path = src/vendor/h2
 	url = https://github.com/aantron/ocaml-h2.git
+[submodule "src/vendor/paf"]
+	path = src/vendor/paf
+	url = https://github.com/dinosaure/paf-le-chien
+	branch = sync-with-dream
--- a/dream.opam
+++ b/dream.opam
@@ -13,7 +13,7 @@ dev-repo: "git+https://github.com/aantron/dream.git"
 
 build: [
   ["dune" "build" "-p"
-    "dream,gluten,gluten-lwt,gluten-lwt-unix,websocketaf,httpaf,httpaf-lwt,httpaf-lwt-unix,hpack,h2,h2-lwt,h2-lwt-unix"
+    "dream,gluten,gluten-lwt,gluten-lwt-unix,websocketaf,httpaf,httpaf-lwt,httpaf-lwt-unix,hpack,h2,h2-lwt,h2-lwt-unix,paf"
     "-j" jobs]
 ]
 
@@ -31,6 +31,7 @@ install: [
   ["opam-installer" "--prefix" prefix "src/vendor/h2/h2.install"]
   ["opam-installer" "--prefix" prefix "src/vendor/h2/h2-lwt.install"]
   ["opam-installer" "--prefix" prefix "src/vendor/h2/h2-lwt-unix.install"]
+  ["opam-installer" "--prefix" prefix "src/vendor/paf/paf.install"]
 ]
 # TODO Set all these packages as conflicts.
 # TODO Use dune install -p ...
@@ -69,6 +70,7 @@ depends: [
   # "gluten-lwt-unix"
   # "httpaf"
   # "httpaf-lwt-unix"
+  # "paf"
 
   # Dependencies of vendored packages.
   "angstrom" {>= "0.14.0"}
--- a/dune-project
+++ b/dune-project
@@ -1,5 +1,5 @@
 (lang dune 2.7)
-(implicit_transitive_deps false)
+(implicit_transitive_deps true)
 
 ; TODO LATER
 ; Note: implicit_transitive_deps is useful during massive development, but is
--- a/example/x-mirage/_tags
+++ b/example/x-mirage/_tags
@@ -1,0 +1,1 @@
+true: package(digestif.c)
--- a/example/x-mirage/config.ml
+++ b/example/x-mirage/config.ml
@@ -1,0 +1,13 @@
+open Mirage
+
+let port =
+  let doc = Key.Arg.info ~doc:"port of HTTP service." [ "p"; "port" ] in
+  Key.(create "port" Arg.(opt (some int) None doc))
+
+let dream =
+  foreign "Unikernel.Make"
+    ~keys:[ Key.abstract port ]
+    ~packages:[ package "dream" ~sublibs:[ "mirage" ] ]
+    (time @-> stackv4v6 @-> job)
+
+let () = register "dream" [ dream $ default_time $ generic_stackv4v6 default_network ]
--- a/example/x-mirage/unikernel.ml
+++ b/example/x-mirage/unikernel.ml
@@ -5,8 +5,13 @@ module Make (Time : Mirage_time.S) (Stack : Mirage_stack.V4V6) = struct
 
   let dream _ = Dream.respond "Hello World!"
 
+  let builtin =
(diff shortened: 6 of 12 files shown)


aantron:
Thanks for the PR and summary!

I gave this PR a first look, and it looks both reasonably structured and surprisingly brief, which is nice.

I plan to finish docs and push out an alpha1 to opam during this weekend, then give this PR a thorough review and merge it. It doesn't look like I'll have to touch anything for alpha1 that this PR also touches, except maybe the `opam` file, but I'll just patch this PR to match any changes to `opam` I make.

aantron (commented):
The PR looks good for starting the Mirage port. Can you separate the Mirage code into an opam package `dream-mirage`? At the moment, with this PR, a user that wants to try out the "ordinary" `dream` is forced to install Mirage.

It looks like little will have to change. You can still use the same internal identifier names that the code is using now. I think `new_app` will have to be prefixed.

aantron:
> Can you separate the Mirage code into an opam package `dream-mirage`?

Or `dream-os` :P

dysinger:
holy burrito, batman. you aren't going to get many contributions by force-pushing master 47 times. :|

aantron:
It's tweaks to README.md and other markdown files immediately after initial large edits, catching typos without adding a huge and pointless stack of commits. And they don't interfere with this PR, nor is anyone cloning the repo in the interim, and if someone is, it is trivial to rebase.

dysinger:
Should we rebase this or was your earlier comment "please put this in another repo?"

aantron:
> Should we rebase this or was your earlier comment "please put this in another repo?"

If you mean the `dream-mirage` package, I originally meant as an second opam package in this same repo. It just needs to be separate, so that we can separate the dependencies of Mirage and non-Mirage versions of Dream (they are, AFAIK, mutually exclusive). Of course, if @dinosaure thinks or becomes convinced that it's easier to handle this project in a separate repo, I'm happy with that too, and I'll happily link to it and/or help upstream it in the future.

As for a rebase, I prefer merging the `master` branch into the PR. I'll squash this PR anyway, so the merge is not a big deal, it's just ephemeral ugliness in the PR branch. The important thing is that it avoids a force-push, which will make incremental review pretty difficult.

It looks like only some package metadata is in conflict, so it shouldn't be too hard. I can offer to do it, @dinosaure, but I'm holding off stepping on your branch for now :) Actually, changing this to be in `dream-mirage.opam` would likely have avoided the `dream.opam` conflict to begin with &mdash; not that it's a big deal anyway, though.

Thanks for the ping, @dysinger!

dinosaure:
Yes, I will try to find some times to reword on that but, with a private discussion with @aantron, I think the best is to keep `dream-mirage` into this repository :+1: - mostly because `dream` vendors some packages used by `dream-mirage`, so to keep a safe zone for both (`dream` and `dream-mirage`), the best is use only one repository.

Then, more concretely, this PR is in stand-by because, into the MirageOS eco-system, we still have few bugs (specially about let's encrypt certificates) and we need to fix them before to burn my limited time on `dream-mirage`. Then, @aantron is aware about my view and MirageOS constraints, so it will take a time but I keep the goal in my head 😃 .

aantron:
> Then, @aantron is aware about my view and MirageOS constraints, so it will take a time but I keep the goal in my head 😃 .

Yes, this PR is only adding initial Mirage-Unix support, and we will then have to refactor things over a small stretch of time to get full MirageOS support. BTW @dinosaure, would you consider changing the PR title about that, for the benefit of other readers?

dinosaure:
This PR is old and another will appear soon with a better integration according to my last work.

PR #42, opened 2021-04-21
Mention Spin to create new projects

tmattio:
This PR recommends Spin to create new Dream projects in the Quickstart section.

I've put the mention of Spin higher than the quickstart script, but of course, I'm happy to change that if you feel the middleware example is a better entry point for new users.

I'm creating the PR as draft as there are a few things I'd like to add to the template before it is recommended. Namely, making the setup of TailwindCSS and Turbolink optional, and adding some boilerplate examples with Caqti.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,16 @@ include their own build commands. They don't have to be subdirectories of
 Especially consider starting with the [full-stack examples][fullstack], which
 build both a Dream server and a JavaScript client.
 
+### Project template
+
+To quickly start a new project with [Spin](https://github.com/tmattio/spin), you can use:
+
+```bash
+opam spin new https://github.com/ocaml-templates/spin-dream.git
+```
+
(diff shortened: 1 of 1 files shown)


aantron:
I would indeed move it lower, after the script but before opam, because Dream has a sort of "opt-in" design, where you start with the minimal and know exactly what you're adding every step of the way.

However, a hypothetical higher-level framework based on Dream might do it the other way. But that higher-level framework might be just a good template :)

aantron:
If it helps to make it more visible, maybe it can get a section heading. Like "### Starter project" or "### Project template"

tmattio:
@aantron I improved the Spin template substantially (all of the features are now optional), and the latest Spin release comes with an Opam plugin, so with Opam install, users can run:
```
opam spin new https://github.com/ocaml-templates/spin-dream.git
```

I'm happy with the state of the workflow, so this is ready to review if you want 🙂

aantron:
Thanks! I opened several issues and PRs in the spin-dream repo.

Issue #3, opened 2021-04-04
Use `multipart_form` instead of `multipart-form-data`


dinosaure:
I know that I come like someone who want to push his projects but I really think that you probably want to use [`multipart_form`](https://github.com/dinosaure/multipart_form) instead of [`multipart-form-data`](https://github.com/cryptosense/multipart-form-data) when the latter does not stream contents - which limits your usage.

If you are interested by that , I can propose a PR if you want :+1:.

aantron:
@dinosaure Thanks! I indeed am interested in `multipart_form`, and have seen it preferred in an Opium discussion when I was first working with file uploads before starting Dream.

I have not yet fully evaluated the two, and chose to start with `multipart-form-data` mainly because its interface seemed easier to deal with quickly for prototyping purposes, since Dream is in an alpha state. Concerning streaming, it appears that `multipart-form-data` does emit chunks, and does accept a stream as the data source. I did not verify yet whether it waits to load its entire input stream, which would, of course, be a major flaw for Dream's purposes, if that is the case. I *did* observe several other flaws (at least that's how they seemed to me):

https://github.com/aantron/dream/blob/f50145650f26cbbd8ce474b0b42570a1c48b940e/src/middleware/upload.ml#L70-L76

I would welcome a PR in principle. However, in the interest of respecting your time, I'd like to delay at first, so that I have a chance to look more closely before asking/agreeing for you to proceed :) I will probably spend a few days finishing up the examples and docs for an "alpha1" release. After that, I will be doing passes through all the parts of Dream, and then I will take many close looks at uploads again :)

dinosaure:
Ok, let me know when you did your choice :) then, if you want to improve some details, I'm able to spend my time to improve the library if you want :+1:.

aantron:
@dinosaure I took a good look at `multipart_form`, and I'd be happy to merge a PR switching to it. The current parser is in [`src/middleware/upload.ml`](https://github.com/aantron/dream/blob/master/src/middleware/upload.ml). It uses a pretty hacky state type which is defined in Dream's core module, because it is a mutable part of requests (the same as the request body stream is inherently mutable):

https://github.com/aantron/dream/blob/54e228d65c4598e1985d8d81504bbeb804dd1de1/src/pure/inmost.ml#L33-L59

[...]

PR #45, opened 2021-04-22
Use multipart_form instead of multipart-form-data

dinosaure:
As mentioned in #3, this PR is a first-shot to be able to stream and upload large files with `dream`. I tried to respect as much as I can the initial interface and tweak a bit some details but:
1) the initial example still works with the new interface
2) I added an other example to upload and save a file at a specific location (which completely ignore security values)

I'm not very convince by the API and the internal state of the request but I don't have the time to figure out about a better interface :confused:. The state uses a trick about `object` to generate unique ID and I think this is the only obscure part of my code.

Then, the "state-machine" is pretty simple and we don't need the ancient plumbing with _streams_ & _tasks_.

Finally, the `multipart_form` package (what is needed for this PR) is unreleased but it's about few functions such as `*.to_string` ot `Header.to_list`. Hope that this PR is good, feel free to share a feedback :slightly_smiling_face: .


tmattio:
Just skimming in, @dinosaure you might want to provide a script to generate the 100MB file (can be in a dune rule to automate it for users) instead of committing it to the repo, otherwise, it'll slow down the git clones for contributors

dinosaure:
> Just skimming in, @dinosaure you might want to provide a script to generate the 100MB file (can be in a dune rule to automate it for users) instead of committing it to the repo, otherwise, it'll slow down the git clones for contributors

Ah yes, I forget about that, sorry.

aantron:
Thanks! I'm reviewing the code now.

I'm initially puzzled by this phrase, though:

> this PR is a first-shot to be able to stream and upload large files with `dream`.

AFAIK the current dependency `multipart-form-data` is also able to both stream and upload large files, and Dream uses it correspondingly (as you must have seen in the code). Can you positively confirm that `multipart-form-data` *cannot* stream? What does this comment refer to? Does `multipart-form-data` do poor flow control, leading to an accumulation of data and growth of internal buffers? Does it buffer large chunks for processing in some way, that is not obvious from the API? Does it have some bugs with streaming? Something else?

src/middleware/upload.ml
```diff
@@ -7,209 +7,118 @@
...
+  | Field.Field (field_name, Field.Content_type, v) ->
+    Lwt.return (`Field ((field_name :> string), Content_type.to_string v))
+  | Field.Field (field_name, Field.Content_disposition, v) ->
+    request.specific.upload.filename <- Content_disposition.filename v ;
+    request.specific.upload.name <- Content_disposition.name v ;
+    Lwt.return (`Field ((field_name :> string), Content_disposition.to_string v))
+  | Field.Field (field_name, Field.Content_encoding, v) ->
+    Lwt.return (`Field ((field_name :> string), Content_encoding.to_string v))
+  | Field.Field (field_name, Field.Field, v) ->
+    Lwt.return (`Field ((field_name :> string), Unstrctrd.to_utf_8_string v))
```
aantron, line 22: AFAICT this changes the meaning of `` `Field`` from "non-file form field" (e.g. corresponding to `<input type="text">`) to "header field`, (e.g. `Content-Disposition: ...`). It looks like, in at least some cases, it is redundant. For example, the code already appears to retrieve the file name and field name.

How does `multipart-form-data` distinguish non-file form fields and file fields? Reading RFC 7578, I see no reliable way to do it. Is there some non-RFC convention that has developed in practice in HTTP?

I want to note that even though the Dream-level API has almost the same constructors as the previous `mutlipart-form-data`-based API, it appears to use them completely differently.

If there is no reliable way to distinguish non-files from files at the protocol level (in practice, regardless of RFC), then it seems that both the `multipart-form-data` and the Dream APIs are wrong. They probably need a fatter `part` type that carries all the headers sent with each part, and can be passed to the application to interpret according to its own knoweldge of the meaning of each part. Is that what you were referring to with:

> I'm not very convince by the API

> I don't have the time to figure out about a better interface

I'll do it in follow-on work, but please do let me know what the problems of the current interface are (without necessarily figuring out a new one :P)

src/middleware/upload.ml
```diff
@@ -7,209 +7,118 @@
...
+      let open Multipart_form in
+      let tree = flatten tree in
+      let fold acc { Multipart_form.header; body= uid; } =
+        let body = List.assoc uid assoc in
+        match Header.content_disposition header with
+        | None -> acc
+        | Some v ->
+          match Content_disposition.name v,
+                Content_disposition.filename v with
+          | Some name, None -> (name, `Value body) :: acc
```
aantron, line 115: This appears incorrect w.r.t. [RFC 7578](https://tools.ietf.org/html/rfc7578#section-4.2):

```
   The file name isn't mandatory
   for cases where the file name isn't available or is meaningless or
   private; this might result, for example, when selection or drag-and-
   drop is used or when the form data content is streamed directly from
   a device.
```

However, again, I don't know if real HTTP clients actually emit no filenames for files under any conditions, or if no filename always comes only from a non-file field. Do you know this?

src/middleware/upload.ml
```diff
@@ -7,209 +7,118 @@
...
+         ; field_to_string request (List.nth header n) )
+    else
+      ( log.debug (fun m -> m "Start to upload contents.")
+      ; Lwt.return (`Part (request.specific.upload.name, request.specific.upload.filename)) )
+
+and upload (request : Dream.request) =
+  if request.specific.upload.uid = Lazy.force Dream.epsilon
+  then
+    let content_type =
+      let ( let* ) = Option.bind in
```
aantron, line 70: I switched Dream to `let%lwt` to preserve users' stack traces as much as possible.

src/dream.mli
```diff
@@ -835,7 +835,7 @@ val form : request -> (string * string) list form_result promise
 (** {2 Upload} *)
 
 type part = [
-  | `Files of (string * string) list
+  | `File of string * string
```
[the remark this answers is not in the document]
dinosaure: With a big picture, we can rollback that and "merge" files with the same name into one and unique `Files`. I did this change because it's more 1-to-1 to what `Multipart_form` gives to us - it returns a list of all parts (even if name of them appear several times).

src/pure/inmost.ml
```diff
@@ -11,41 +11,36 @@ include Status
...
-  mutable chunk_listener : string option Lwt.u option;
-  mutable last_field_name : string option;
-  mutable last_file_name : string option;
-  mutable buffered_chunk : string option;
-  mutable next_file : bool;
-  mutable continue : unit Lwt.u;
-  mutable fields : bool;
-  mutable field : unit -> (string * string) option Lwt.t;
-}
+type uid = < >
```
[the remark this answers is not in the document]
dinosaure: `Multipart_form` asks to "identify" a part because it considers that a stream of a part can be consumed concurrently with the parsing of the whole body. The idea is to use this unique identifier to identify a part and to bind it with a `Leaf` of the returned `Multipart_form.t`.

With such a bind, we are able to retrieve the stream of the part from the unique ID and the part (including its header) from the _tree_ `Multipart_form.t`. Generally, this unique ID can be an integer or something else which derives from the given `Header.t`. This how `multipart_form` was designed.

In this specific context, the `uid` is used to differentiate 2 states when we process the `multipart/form-data`:
- when we start the process (`uid = Dream.epsilon`)
- when we manipulate a part

In a certain sense, `uid` is like the old `state.initial` value used before this PR. If you want to rollback to such value, I can propose something 👍 .

src/middleware/upload.ml
dinosaure: From my point of view, a `multipart/form-data` (according to RFCs) is a tree of parts (leafs) where you can find for each of them several HTTP fields (such as `Content-Disposition` or `Content-Transfer-Encoding`). So the current behavior of:
```ocaml
val upload : request -> [ Field of string * string | Part of (string option * string option ]
```

Is to deliver to the end-user fields of the the current part as a _stream_ such as:
```ocaml
# upload request ;;
- [> `Field of string * string ] = `Field ("content-type", "text/html")
# upload request ;;
- [> `Field of string * string ] = `Field ("content-disposition", ...)
# upload request ;;
- [> `Field of string * string ] = `Field ("content-transfer-encoding", "base64")
# upload request ;;
- [> `Part of string option * string option ] = `Part (possible_name, possible_filename)
# upload_part request ;;
# upload request ;;
- [> `Field of string * string ] = `Field ("content-type", "image/png")
...
```

It's like a stream of flattened parts. So we don't dissociate files and values. We just say that we have a part with a possible `name` and a possible `filename` extracted from a possible `Content-Disposition`. Note that the part can contains several others fields (like `Comment` or extra information). Even if I did not notice that, it is allowed by the RFC.

src/middleware/upload.ml
dinosaure: Depending on what you want. For my perspective, `multipart-form` does not dissociate values (from `<input ...>`) and files (from `<input type="file" ...>`) and manipulate just _parts_. According to the RFC, we don't have any real clues to say that this specific part is a file or a value. Only the `filename` parameter can help us and it's not a mandatory (as you said) parameter.

So I'm not really sure about these lines but they want to go further (than the RFC) on the interpretation of parts.

dinosaure:
> AFAIK the current dependency multipart-form-data is also able to both stream and upload large files, and Dream uses it correspondingly (as you must have seen in the code). Can you positively confirm that multipart-form-data cannot stream? What does this comment refer to? Does multipart-form-data do poor flow control, leading to an accumulation of data and growth of internal buffers? Does it buffer large chunks for processing in some way, that is not obvious from the API? Does it have some bugs with streaming? Something else?

So `multipart-form-data` can stream _out_ the output but I think it does that in the wrong way. From experimenting with @Armael on `multipart_form`, we noticed several issues about `Lwt_stream` and `lwt` in general. We did the choice to use `Lwt_stream.create_bounded` internally because it seems that `lwt` does not take the opportunity to reschedule properly between the reader and the writer (even if the _reader_ is plugged to a _waiting_ _syscall_ - such as `read`).

We noticed that we don't let any chance to OCaml to _garbage-collect_ previous chunks and this is why we decided to use `Lwt_stream.create_bounded` to blocks and:
1) reschedule to the writer
2) clean old _chunks_

We tried to play with `Lwt.pause ()` and even if it's better, it's less deterministic than a use of `Lwt_stream.create_bounded`. Then, by design, I really think that we should let the opportunity to use `Lwt.async`/`Lwt.both` to the end-user to show that we really able to concurrently execute the consumer & the producer. `multipart-form-data` does not really highlight such usage - and I think it's why you wrote such plumbing before.

Moreover, `multipart_form` comes from a huge work about RFCs where RFC 7578 is only the top. The bedrock of all of these stuffs is the [RFC 822][rfc822] unfortunately which can allow many (many ...) surprises. For example, we can talk about the `FWS` token which allows to write a field value into multiple lines (and, as far as I can see, `multipart-form-data` does not handle such case - which is pretty common if it's combined with the limit of 80 characters per line). So I developed [`unstrctrd`][unstrctrd] to takes care about this "detail".

RFC 7578 deprecates `Content-Transfer-Encoding`, however we can have such way to encode values. It's specially true for the `quoted-printable` encoding which is handled by [`pecu`][pecu]. The base64 encoding exists too and in a different way than what we usally think about base64. It's why we extended [`ocaml-base64`][ocaml-base64] to handle [RFC 2045][rfc2045] contents. All of that are behind `multipart_form` - you don't need to deal with quoted-printable or base64.

I think, if we want to compare multiple implementations of `multipart/form-data`, we should take about such details. In most of our cases, all of these details don't really appear for an usual case but I already discovered these mysteries in the web. I know that we have a huge gap between what the RFC describes and what we commonly use - and thanks to @Armael to help to fill this gap. But I'm convince that the right way to handle is not to discard historic RFCs which are the part of the result by (historic) inheritance.

[rfc822]: https://tools.ietf.org/html/rfc822
[unstrctrd]: https://github.com/dinosaure/unstrctrd
[pecu]: https://github.com/mirage/pecu
[ocaml-base64]: https://github.com/mirage/ocaml-base64/pull/19
[rfc2045]: https://tools.ietf.org/html/rfc2045

src/dream.mli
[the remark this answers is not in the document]
aantron: > "merge" files with the same name into one and unique `Files`. I did this change because it's more 1-to-1 to what `Multipart_form` gives to us - it returns a list of all parts (even if name of them appear several times).

This reads as if it is about merging files with the same filename into one file. But yes, I mean merging files with the same *field* name into a list of files for that field.

`multipart/form-data` is ultimately an implementation detail from the point of view of a framework user. What they see is that they emitted a form like

```html
<form ...>
  <input type="text" name="foo">
  <input type="file" name="bar" multiple>
</form>
```

and in the easy-to-use buffering helper, it should be obvious how to retrieve the whole value for each one of those `<input>` fields, without any boilerplate that will have to be repeated everywhere one wants to do so. So I think the files should be merged into a list (as they already were).

I also imagine that if we checked, we would find that all user agents and clients send the file parts for `<input multiple>` contiguously anyway, so it's not an unreasonable representation of a real `multipart/form-data` transmission to represent these parts as a list, either. Of course, someone could probably use curl to send parts completely out of order, but there is no reason AFAIK why we should represent that directly in the easy-to-use API.

src/pure/inmost.ml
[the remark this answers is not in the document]
aantron: > that a stream of a part can be consumed concurrently with the parsing of the whole body.

Does this mean something like, you are effectively parsing a node of a multipart tree, while directly parsing one of its leaves? Because otherwise I don't see how *one* body can be consumed concurrently (with itself?), since the parts and their whole structure arrives serialized.

Can the uid be just an integer? My guess is that the uid has to be unique within one message body. If that's right, I think, even on a 32-bit system, we are unlikely to have so many concurrent multipart parsers, that an id could get incremented 2^31 times and wrap all the way around to a collision within a single stream. Of course, that leaves the question of what to use for epsilon. I hadn't even realized that such a value would be necessary. Maybe it's possible to increment by 2 and use an odd number for epsilon :D Then we have 2^30 ids margin for collision.

> In a certain sense, `uid` is like the old `state.initial` value used before this PR. If you want to rollback to such value, I can propose something 👍 .

It's not necessary to roll back. I just don't see the motivation for specifically this kind of API, yet :P However, I will accept it anyway, but I want to see if there is something less mysterious to future readers (including me) that objects :) And hopefully doesn't require conditional code.

src/middleware/upload.ml
aantron: Ok, after doing some tests, I think `` `Value`` should be removed from the API entirely. Even though browsers don't seem to generate parts without `filename` for files, and it is difficult (impossible?) to do with curl even when reading STDIN, it *is* possible to use curl to submit a field that has file content inlined. Then, it won't have a filename.

Regardless, it is easier for the user, probably, not to deal with two different cases `` `File`` and `Value`. Apps can decide which field should be interpreted as file(s) based only on the field name. The presence, absence, or value of `filename` probably should not be trusted at all, in general &mdash; it is probably only good for informative purposes, to identify files back to the user.

If you're willing, remove it in this PR :) It's also fine to leave it for me to do later.

I added an example, [**`w-multipart-dump`**](https://github.com/aantron/dream/tree/master/example/w-multipart-dump#files), for inspecting `multipart/form-data` bodies sent by clients (as an aside, it appears CSRF checking is not being triggered in it &mdash; will look into that after this PR is settled).

aantron: I also noticed that providing zero files causes at least Chrome and Firefox to send

```
-----------------------------250068058125696789421882203568
Content-Disposition: form-data; name="files"; filename=""
Content-Type: application/octet-stream


-----------------------------250068058125696789421882203568--
```

i.e., what appears to be a file with an empty filename and no body. This looks like a complication. I think it's fine not to address that in this PR. It should probably be emitted like that in the streaming API, with a note in the docs. For the easy-to-use API, it should probably be filtered out, also with a note in the docs that this is being done.

dinosaure: What do you think about:
```ocaml
type part = { filename : string option; contents : string; }

val multipart : request -> [ `Ok of (string * part) list | ... ]
```

aantron: That looks right.

Is there any point in trying to emit the per-part headers? AFAIK they are all untrustworthy under "wild" HTTP conditions, though maybe they are useful in closed systems somehow? It seems they are obscure.

src/pure/inmost.ml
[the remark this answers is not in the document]
dinosaure: I deleted the usage of `type uid = < >` and choose to have something more _readable_ for a maintainer. A state still is needed to separate the initial operation (launch the consumer stream), the header state to emit `Field` and the part state to emit contents. I added an explanation about what is going on 👍 .

src/middleware/upload.ml
dinosaure: Deleted any use of `let*` in my codebase.

dinosaure:
I pushed-force a new version (without any changes on examples due to a cryptic compilation error with `ppx` - feel free to push on my branch to fix it if you want). The internal state to implement `upload` and `upload_part` is much more easier (about maintenance - in my view) and I merged values with the same `name` to correctly handle multiple files.

Let me know if something is not good for you :+1:.

src/middleware/upload.ml
dinosaure: Yes, I will try to extend the record with an `headers` fields 👍

aantron:
Just to put down in writing here messages from chat, the cryptic errors are caused by https://github.com/ocaml-ppx/ppxlib/issues/221, and it should be possible to fix them, for the time being, by downgrading ppxlib to 0.20.0. Once there is a fix, I'll have Dream require the fixed version, so that this doesn't come up again (hopefully!) (#56).

dinosaure:
I forced-pushed the last version which includes `headers` field into the `part` type. I'm not a good writer for documentation (comparing to you) but I tried to explain on my way (in `examples/w-stream-upload`) what is going on. I think, this PR is mostly done regardless your feedback :+1:. So if you have some concerns, feel free to tell me.

PS: the fix about `ppxlib.0.20.0.` works, thanks for that :+1:.

aantron:
Thank you!

PR #52, opened 2021-04-28
Add hot reloading example

tmattio:
Inspired by `dream-serve`, this PR adds an example that demonstrates how to set up hot reloading with Dream.

The WebSocket logic is a bit different from `dream-serve` since the assumption here is that the client should reload when the server restarts, that is, when the WebSocket connection is lost. To achieve this, once the connection is lost on the client side, it will try to reconnect to the server. Once it is able to reconnect, it triggers a page reload.

The example documentation can certainly be improved, I'll re-read it with a fresh view tomorrow, but early feedback is appreciated.

I'm also curious how you'd feel about having the `hotreload_script`, `inject_hotreload_script` and `hotreload_route` functions (or any other API for hot reloading) directly in Dream? I figure hot reloading is a feature that most people will want to benefit from by default.


example/w-hot-reloading/hot_reloading.ml
```diff
@@ -0,0 +1,89 @@
...
+let inject_hotreload_script
+    ?(reload_script = hotreload_script ())
+    ()
+    (next_handler : Dream.request -> Dream.response Lwt.t)
+    (request : Dream.request)
+    : Dream.response Lwt.t
+  =
+  let%lwt response = next_handler request in
+  match Dream.header "Content-Type" response with
+  | Some "text/html" | Some "text/html; charset=utf-8" ->
```
tmattio, line 59: Checking if it starts with `text/html` is probably more robust

example/w-hot-reloading/hot_reloading.ml
```diff
@@ -0,0 +1,89 @@
...
+    let%lwt body = Dream.body response in
+    let soup = Soup.parse body in
+    let open Soup.Infix in
+    (match soup $? "head" with
```
tmattio, line 63: Would be nice to drop the dependency to `lambdasoup`. Would an ad-hoc search for `</head>` work? 🤔

example/w-hot-reloading/hot_reloading.ml
```diff
@@ -0,0 +1,89 @@
...
+};
+|js}
+    route
+    retry_interval_ms
+    max_retry_ms
+
+let inject_hotreload_script
+    ?(reload_script = hotreload_script ())
+    ()
+    (next_handler : Dream.request -> Dream.response Lwt.t)
```
tmattio, line 53: Drop the types, too noisy

example/w-hot-reloading/esy.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies": {
+    "@opam/dream": "aantron/dream:dream.opam",
+    "@opam/dune": "^2.0",
```
aantron, line 4: This also needs `"@opam/lambdasoup": "*"`.

example/w-hot-reloading/hot_reloading.ml
```diff
@@ -0,0 +1,89 @@
...
+    ?(route = "/_hotreload")
+    ()
+  =
+  Printf.sprintf
+    {js|
+var socketUrl = "ws://" + location.host + "%s"
+var s = new WebSocket(socketUrl);
+
+s.onopen = function(even) {
+  console.log("WebSocket connection open.");
```
aantron, line 13: Not familiar with developer expectations for this, but do we need logging in such a script? Personally, I would try to observe the activity from the Network tab, which I assume most browsers offer.

aantron (commented):
Great! This is exactly the sort of thing I was imagining would be good for example `w-auto-reload` in the [example roadmap](https://github.com/aantron/dream/tree/master/example#roadmap), and also in the [dream-serve roadmap](https://github.com/aantron/dream-serve#roadmap).

Regarding the potential Lambda Soup dependency if upstreaming this to Dream, that's one of the reasons I was instead considering putting it in dream-serve (or yet another project). dream-serve already depends on Lambda Soup, and does something quite similar, so it seems like it should be natural to package it as a library and an executable, where the library essentially contains a somewhat elaborated version of this code.

OTOH Lambda Soup doesn't really have any heavy dependencies of its own, so maybe it's justified to depend on it, the same way as Dream already depends on Caqti. It will just take a bit more compilation time during dependency install.

Regarding the example text, another option is that I can merge it as is. I will probably tweak it regardless, so I could just work from a draft, sketch, or even an outline of what you'd like to be there, and it looks like you've already got a good draft.

aantron:
Also, I suspect this is actually live reloading and not hot reloading (https://github.com/aantron/dream-serve/issues/2), and both this example and the description of dream-serve will need to be changed.

example/w-hot-reloading/hot_reloading.ml
aantron: An ad-hoc search for `<head>` seems better, as there is less chance to end up in a `<script>` tag or similar, and match a quoted `</head>`. However, not all HTML actually has a literal `<head>` element when transmitted. That includes most of Dream's examples. For instance, in [**`1-hello`**](https://github.com/aantron/dream/tree/master/example/1-hello#readme):

```
let () =
  Dream.run (fun _ ->
    Dream.html "Good morning, world!")
```

The `<head>` tag is automatically inserted by HTML5-conforming parsers when loading such a document, though.

Given that developers are probably likely to start from HTML that doesn't have an explicit `<head>`, it seems like scanning for it will tend to bite them.

example/w-hot-reloading/hot_reloading.ml
aantron: It should probably be a check that it either

- Equals `text/html`.
- Contains `;`, and the text before the first `;` equals `text/html`.

tmattio:
> Also, I suspect this is actually live reloading and not hot reloading

I learned something! https://stackoverflow.com/questions/41428954/what-is-the-difference-between-hot-reloading-and-live-reloading-in-react-native!

Thanks 😄

aantron:
Thanks for posting that link. I updated the description of dream-serve :)

example/w-hot-reloading/hot_reloading.ml
tmattio: I'm not sure either.. Personally, I feel reassured to have the logs tell me that everything is going well, but that might only be because I wrote the script.

The log is not useful in its current state though. It could be something like "Live-reload WebSocket connection established, waiting for the server to restart to reload the page."?

example/w-hot-reloading/hot_reloading.ml
tmattio: Right, what we could do also is locate the last `</body>` tag and insert the script before it? I think it would be resilient to scripts containing this tag. And if we don't find it, we can simply add the script at the end of the HTML document?

tmattio:
> that's one of the reasons I was instead considering putting it in dream-serve 

I thought about it too, but `dream-serve`'s behavior is a bit different since it answers the client itself. Are you thinking of having a "proxy" mode in `dream-serve`? To answer on `_websocket_monitoring` and forward the rest of the requests to the given port/interface?

That also seems fine to me, I don't have any preferences. The only drawback I would see maybe is the duplication of settings in this setup, where you'd configure your app to listen on a specific port and you'd also have to configure `dream-serve` to forward to this same port. But that's a very small drawback..

aantron:
> I thought about it too, but `dream-serve`'s behavior is a bit different since it answers the client itself. Are you thinking of having a "proxy" mode in `dream-serve`? To answer on `_websocket_monitoring` and forward the rest of the requests to the given port/interface?

No, I'm thiking that dream-serve will become primarily a library, which will contain a handler and/or a middleware for live reloading. The dream-serve package will also include a dream-serve executable, which will use that library to build a static site server. Users who want to add live reloading to their dynamic Web app will depend on the package but use the library. So it won't be a proxy in either case.

tmattio:
> No, I'm thiking that dream-serve will become primarily a library, which will contain a handler and/or a middleware for live reloading.

I see! In this case, what do you think of having this library in Dream (e.g. `dream.livereload`)? Live-reloading and serving static assets are two different use cases and, as I mentioned, I think live reloading will be needed for most application development. There's also the fact there won't be much code in common between the live-reloading feature and the static serving one, and the dependencies for the latter are not required for the former (`luv` and presumably `lambdasoup` following our discussion above).

Anyways, that's probably outside of the scope of this PR, unless you'd like me to extract the middleware/routes in a library as part of it.

aantron:
Agreed. I'm ready to merge this PR whenver you are done making changes. I can address any nits afterwards and/or expand the README.

example/w-hot-reloading/hot_reloading.ml
cdaringe: 🤷 . if it's there, perhaps it should be more specific, that it's the hot-reload socket, vs any general socket connection :)

example/w-hot-reloading/hot_reloading.ml
cdaringe: this classic problem. here's merely a suggestion:

- add a document rendering middleware, where
  - m/w accepts a mutable map of `request:document-segments`, e.g. `{ req1ref: Some { head: []  }; req2ref: None }`
- some hot-reloading-aware middleware pushes script datas onto `doc_segments[reqXref].head`
- on middleware exit, `Dream.html <html><head>{% ... map doc_segments to_script_tag %} ...`

essentially emulate closures via a mut ref, then don't expect the leafiest middleware to own the final render. instead, the generally leafier middlewares mutate state ( 😢 ), or accumulate state, and on-pop, the doc middleware actually calls `Dream.html`

next.js uses a `<Head />` component that allows participation in document generation, specifically to avoid cases like this, where a parse and re-generate is needed. it's even more quirky in that the component needs know references to the request of which it's modifying, and i think they plan on ejecting from it. regardless, their flow looks like this:

```
HTTP GET /  ->                    _app m/w   ->   react-render
(stream)        <-  (html doc string)         <- (body string)
```

aantron: It's possible to to associate HTML document trees with responses using [request-/response-local variables](https://aantron.github.io/dream/#variables), with something like:

```ocaml
let html_tree = Dream.new_local ()

(* In your handler, or a helper to be called from handlers. *)
response
|> Dream.with_local html_tree my_dom_alike

(* Retrieving it in a middleware. *)
match Dream.local html_tree response with
| Some dom_alike -> Dream.html (render dom_alike)
| None -> (* ... *)
```

I'm not sure this is worth it for a hot-reloading middleware &mdash; I assume it would be used in development only, and parsing the HTML again is, AFAIK, not an issue. However, this could be useful in a higher-level framework that needs to defer HTML rendering in production.

example/w-hot-reloading/hot_reloading.ml
aantron: I still favor removing it, as:

- This is an example, and it's best to keep it brief, to make it easier to focus on the actual machinery.
- The information is redundant with the network tab, and direct observation that reloading continues to work.

tmattio:
Ok, feel free to merge whenever you want! If you'll like, I'll land another PR to extract the live reloading feature in a library as we discussed.

aantron:
Thanks! Looking forward to the library. I'll be happy to link it from the README.

PR #53, opened 2021-04-28
Add welcome page and error templates

tmattio:
This gets the ball rolling on the welcome page and a better error template.

As I mentioned on Discord, the welcome page is adapted from TailwindCSS's old playground, so I'd like to re-write it before it's merged, but that's a detail and we can discuss the other parts of the PR independently 🙂 

I modified the `hello` example to serve the welcome page, so the PR can be tested with `dune exec example/1-hello/hello.exe`.

I've also updated the default error template.

PS: I'm opening multiple PRs, but don't feel any rush to review/merge, I know you have your own todo list for Dream's first release 😉


aantron:
I am very happy to have a nice default error template.

Regarding a welcome page:

- I would like to keep the examples very clear &mdash; part of the goal of Dream is to be as comprehensible as possible to the developer. `1-hello` should show that Dream is an interaction between `Dream.run` and a plain function, and replacing it by a template obscures that.
- In general, I don't see where it would be useful in Dream itself. There is never really an initial state in a base Dream project, AFAIK, where Dream would have an opportunity to show a welcome page. It is likely very useful for project generators and the like. I don't know, this early, if all project templates/generators would want to offer a default error page. It seems that if they have different combinations of technologies, they would want to both use those technologies for their welcome pages, and link to them from the same pages.

One place where a nice welcome page would be *very* useful in this repo is the online [playground](https://github.com/aantron/dream/tree/master/example/z-playground#files). The prototype is currently running at [http://dream.as](http://dream.as):

Снимок

As you can see, it has a welcome template rendered by the default example, which is very 1993 due to visual space limitations in the code. It would be very nice to add a welcome template to the playground in a library that would be loaded there, that can just be called by the starter code. I already had to drop several things from it because they were cluttering the editor in the left pane.

tmattio:
> In general, I don't see where it would be useful in Dream itself

I agree, it doesn't seem useful to expose it in the API. However, I can see a use case to have an official Dream welcome page: it does make sense to me that the playground and project generators such as Spin share the welcome page. In this case, what do you think of putting this in a `dream.welcome` library that `dream` does not depend on?

aantron (commented):
> it does make sense to me that the playground and project generators such as Spin share the welcome page.

Why? I think the playground will want to link, for instance, to a list of examples that are pre-loaded into the playground. On the other hand, I think a good welcome page for a Spin template will link to docs for the additional things included in the template, such as Tailwind, Turbolink, etc. (even if some of those things are gradually getting upstreamed). In addition, while the playground template can be implemented in any way (it is opaque), a project template's (or higher-level framework's) welcome page should probably at least minimally showcase the combination of libraries/frameworks used by that template. So, I imagine that all these pages will be different. They are really not welcome pages for Dream then, but for different projects based on Dream.



> what do you think of putting this in a `dream.welcome` library that `dream` does not depend on?

Given the above, I think the playground should get a template in a `playground` or `dream-playground` library (which will not be published), and we will link that library into all playground sandboxes.

src/http/fallback_template.eml.ml
```diff
@@ -0,0 +1,81 @@
...
+    </body>
+  </html>
+
+let render_debug ~debug_dump ~code ~reason = 
+  <!DOCTYPE html>
+  <html lang="en">
+    <head>
+      <meta charset="utf-8" />
+      <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no" />
+      <link rel="icon" type="image/png" sizes="16x16" href="data:image/png;base64,<%s! favicon %>" />
```
aantron, line 39: This favicon will not match any application's icon. If we just don't set anything here, the browser should default to `/favicon.ico`, which will be consistent with the rest of the user's application.

src/http/fallback_template.eml.ml
```diff
@@ -0,0 +1,81 @@
...
+      <link rel="stylesheet" href="https://rsms.me/inter/inter.css" />
```
aantron, line 40: We should not make external HTTP requests in the error template of Dream &mdash; it's a privacy violation, amounts to telemetry on developers, and will not work when developing offline (giving at least inconsistency in presentation). This should be inlined with a bundler or data URL, or just omitted, I think. Is it necessary?

src/http/error_handler.ml
```diff
@@ -189,18 +189,22 @@ let customize template (error : Error.error) =
...
-    | None -> response
-    | Some info ->
-      response
-      |> Dream.with_body info
-      |> Dream.with_header "Content-Type" "text/plain"
+    | Some s ->
+      let debug_dump = Dream__pure.Formats.html_escape s in
+      Fallback_template.render_debug ~debug_dump ~code ~reason
+    | None ->
+      Fallback_template.render ~code ~reason
```
aantron, line 202: This changes the behavior described in [*Errors*](https://aantron.github.io/dream/#errors):

> The default error handler logs errors and its template generates completely empty responses, to avoid internationalization issues.

Without `~debug:true`, Dream should not assume any natural language or any style of writing or visual presentation, and leave that completely up to Dream's user (or a higher-level framework or project generator). Otherwise, it adds a new thing that people have to remember to "tuck in" to avoid leaking non-internationalized strings, dealing with style issues, etc.

Conveniently, at least Chromium-based browsers already internationalize their built-in error pages for empty responses, so "base" Dream is piggy-backing on that.

src/http/fallback_template.eml.ml
```diff
@@ -0,0 +1,81 @@
...
+      <style>
+        <%s! css %>
+      </style>
+      <title><%i code %> <%s reason %></title>
+    </head>
+
+    <body class="antialiased">
```
aantron, line 47: Given how relatively simple the error template is, can you remove Tailwind in favor of just plain CSS? The CSS can probably just be written inline, it's quite brief. It looks like together with not overriding the app's favicon, that will simplify away the `asset` directory.

src/http/error_handler.ml
tmattio: Agree! I'm wondering why the error handling is not provided as a middleware then? It seems to offer the same intended flexibility to the user, while re-using existing concepts instead of another semantics?

I'm guessing this is because of how Dream handles errors internally, so maybe that's simply not possible?

tmattio:
> They are really not welcome pages for Dream then, but for different projects based on Dream.

Ok, I'm convinced 🙂 I'll remove the welcome page from this PR then and keep only the error template. I can open another PR to change the welcome page of the playground, but as you suggested, it should be different from the one in this PR.

src/http/fallback_template.eml.ml
tmattio: Not strictly necessary, but Inter font is much better visually than the default system fonts IMO.
I don't think it's possible to inline it though, Inter's CSS itself links a bunch of font files. So I'll remove it.

src/http/error_handler.ml
aantron: > I'm wondering why the error handling is not provided as a middleware then? It seems to offer the same intended flexibility to the user, while re-using existing concepts instead of another semantics?

Error handlers are indeed very middleware-like. But middleware-based error handlers are an (1) opt-in way to (2) handle errors from *within* your app. Dream's error handler handles errors from the app and from the low-level libraries, to capture errors from all levels and pass them through a single handler as a filter. That also means that a default error handler must always be present, because low-level errors (such as 400 Bad Request responses) get generated by the libraries under Dream no matter what an app does or does not do. It would be strange if a 400 Bad Request coming from deep inside http/af got somehow routed into a middleware halfway through a user app's middleware stack for postprocessing by a template. The other option, to have the framework simply emit the low-level 400 Bad Request, leaks undecorated, untranslated strings to the client, that are under the control of http/af, etc., and not under the control of the framework user, the Web app developer.

If there is a TLS handshake error, how would an advanced user gain control over its logging, if error handlers are middlewares?

There is also the issue of what happens when the error handler itself raises an error. If error handlers are hidden in middleware stacks, do we know if any given error is a real error, or a "double fault" coming from a failed error handler? If the framework distinguishes the error handler, it can detect double faults exactly in an easy way, at the call site.

So error handlers are some kind of middleware-like built-in thing that have some slightly different semantics.



> I'm guessing this is because of how Dream handles errors internally, so maybe that's simply not possible?

No, it's not a limitation. Dream could offer middleware-based error handlers as the default, but they would have all the drawbacks above. A user can still implement a middleware-based error handler trivially for their own purposes, if they need to deal with app-only errors in a way that middleware-based error handlers are well-suited for.

src/http/fallback_template.eml.ml
aantron: In practice, you could probably just include the `.woff2` file only, and let very old browsers choose an uglier font. However, I do think it's best to remove it, since I don't think there is a legibility issue with the five or so lines of text that get rendered in the error template in that font, only one of which is variable :) It is much more of a potential issue for the welcome page, but, as per the other discussions, welcome pages really belong in each Dream-based starter project anyway. And I am all for experimenting with fonts in those pages, or other defaults of those projects :)

src/http/fallback_template.eml.ml
aantron: (Realistically, many modern sites might not have a `/favicon.ico`, however, I think it's better to serve nothing here than a hardcoded icon, as then it's equally inconsistent, and I am arguing for using this template for `~debug:true` only elsewhre).

src/http/error_handler.ml
aantron: (By the way, there is a built-in error handling middleware, [`Dream.catch`](https://aantron.github.io/dream/#val-catch), but its job is just to forward app errors to the "big central error handler" that is also catching http/af errors, TLS errors, etc. This middleware can be [removed](https://aantron.github.io/dream/#builtin) to suppress its behavior, for completely custom app error handling. Lower-level errors still won't go into the app, but to the big error handler.)

tmattio:
@aantron I removed the welcome pages and I think I've addressed all of your comments. I also inlined the CSS to remove the `asset` library. I kept the Tailwind classes though: with the CSS reset that's generated with the Tailwind classes, it's not trivial to replicate the style (replacing the classes by their CSS rules for the entire document will not give the same style). I can work on that a bit more if it's a blocker though.

aantron:
Thanks!

I pushed this up to the playground, where you can see it in action at http://dream.as/8-debug:

8-debug-2

<br>

> I can work on that a bit more if it's a blocker though.

Not a blocker :)

src/http/error_handler.ml
```diff
@@ -189,18 +189,21 @@ let customize template (error : Error.error) =
...
-
-  Lwt.return response
+  let status = Dream.status response in
+  let code = Dream.status_to_int status
+  and reason = Dream.status_to_string status in
+  match debug_dump with
+  | None -> Lwt.return response
+  | Some s ->
+    let body =
+      let debug_dump = Dream__pure.Formats.html_escape s in
```
aantron, line 200: I just noticed that this is probably redundant, as `<%s ... %>` inside the template already applies `html_escape`. I'll just patch it in a future 1-line (or 1-char?) commit.

aantron:
@tmattio I've moved the playground welcome template into a sort of runtime library for the playground:

https://github.com/aantron/dream/blob/a2c07622f1000c96f04aaf23c78a05d8bd9c385b/example/z-playground/runtime/playground.eml.ml#L8-L32

If you'd still like to contribute a template, this should be a good place for it!

PR #57, opened 2021-05-06
Improve docker examples

tmattio:
The Docker examples are amazing and really useful. It's rare to see libraries documenting the deployment part, so your efforts on these are really appreciated 🙂 

I thought I'd propose some improvements to the example. I've been using Docker for developing and deploying Opium apps with Docker quite a lot, so hopefully, the setup here should be relatively real-world-proof.

- Building the binaries can be done in Docker instead of copying the files over to a machine compatible with the Docker image. This unlocks one of the main benefits of Docker: the images can be built by any developer, independently of their operating system. To do this, the Dockerfile now contains 2 stages: a stage for the build context where we install the build dependencies and copy the source code in the Docker context. The second stage is a very minimal alpine image where we only install the runtime dependencies and copy the binary produced in the first stage.
- The logging driver in `docker-compose.yml` defaults to `json-file` to be compatible to OSX (`journald` is not available there), and we read the env var `LOGGING_DRIVER` to make sure we use `journald` in the `deploy.sh` script.
- I added a variant of the `z-docker` example for Opam: the Dockerfile is a bit different, and I figured having deployement example is something Opam users will need as well


example/z-docker-opam/.dockerignore
```diff
@@ -0,0 +1,39 @@
+# ocamlbuild working directory
+_build/
+_esy/
+
+# ocamlbuild targets
+*.byte
+*.native
```
aantron, line 7: IMO it's best to skip this. People won't be using Ocamlbuild in their own build, realistically. The same goes for `_esy` for an OCaml project. Also, for the sake of an example (so, to focus on only the most important parts), I wouldn't bother excluding `.merlin` and the like. I would probably even recommend a `.dockerignore` that excludes *eveything*, but then opts back into e.g. `src/` with `!`. In any case, I think the `.dockerignore` should become smaller for an example.

example/z-docker-opam/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+# Install system dependencies
+# Some of these are only needed for database and JavaScript stuff, but this is 
+# our build context, it won't impact our runtime environment.
+RUN sudo apk add --update \
+    make libffi-dev linux-headers m4 ncurses-dev perl pkgconf postgresql-dev \
+    musl-dev git pkgconfig gmp-dev libc-dev openblas-dev zlib-dev \
+    npm nodejs libev-dev
+
+RUN eval $(opam env) && opam repository set-url default https://opam.ocaml.org && opam update
```
aantron, line 11: Is the explicit `set-url` necessary?

example/z-docker-opam/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+# Install dependencies
+# We only copy the opam file to install dependencies only when they change, and 
+# not every time another file is modified.
+ADD hello.opam hello.opam
+RUN eval $(opam env) && opam install . --deps-only
+
+# Build project
+ADD . .
+
+RUN opam exec -- dune build --profile release
```
aantron, line 24: There is no need for `--profile release` AFAIK &mdash; it just stops treating warnings as errors. If someone wants that, they can add it later, but it's not a big need for leaf projects AFAIK.

example/z-docker/esy.json
```diff
@@ -9,7 +9,7 @@
   },
   "esy": {
     "build": [
-      "dune build --root . ./app.exe"
+      "dune build --root . --profile release ./app.exe"
```
aantron, line 12: What is the purpose of this `--profile release`?

example/z-docker/esy.json
```diff
@@ -9,7 +9,7 @@
...
     ]
   }
-}
+}
```
aantron, line 15: Please restore the newline.

example/z-docker/Dockerfile
```diff
@@ -1,4 +1,40 @@
...
+
+RUN eval $(opam env) && opam repository set-url default https://opam.ocaml.org && opam update
+
+WORKDIR /build
+
+RUN npm install esy
+
+# Install dependencies
+# We only copy the opam file to install dependencies only when they change, and 
+# not every time another file is modified.
```
aantron, line 23: This comment is out of sync with the build system being used here. Can you customize this file to `esy`?

example/z-docker/Dockerfile
```diff
@@ -1,4 +1,40 @@
...
+RUN sudo apk add --update \
+    make libffi-dev linux-headers m4 ncurses-dev perl pkgconf postgresql-dev \
+    musl-dev git pkgconfig gmp-dev libc-dev openblas-dev zlib-dev \
+    npm nodejs libev-dev 
+
+# Install Esy-specific dependencies
+RUN sudo apk add \
+    perl-utils
+
+RUN eval $(opam env) && opam repository set-url default https://opam.ocaml.org && opam update
```
ulrikstrid, line 15: Why is this needed when we use esy?

example/z-docker/Dockerfile
```diff
@@ -1,4 +1,40 @@
...
-ENTRYPOINT /app.exe
+FROM ocaml/opam:alpine as build
+
+# Install system dependencies
+# Some of these are only needed for database and JavaScript stuff, but this is 
+# our build context, it won't impact our runtime environment.
+RUN sudo apk add --update \
+    make libffi-dev linux-headers m4 ncurses-dev perl pkgconf postgresql-dev \
+    musl-dev git pkgconfig gmp-dev libc-dev openblas-dev zlib-dev \
+    npm nodejs libev-dev
```
aantron, line 9: AFAIK, many of these are not necessary for an esy build. Have you confirmed that they are? See the instructions in the README.md for an approximate set of needed system dependencies.

example/z-docker-opam/Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+
+FROM alpine:3.12 as run
```
ulrikstrid, line 26: alpine quite often have some pretty bad CVE's that go un-patched. So a debian-slim variant is often a better pick.

example/z-docker/.dockerignore
```diff
@@ -1,4 +1,39 @@
+# ocamlbuild working directory
 _build/
 _esy/
+
+# ocamlbuild targets
+*.byte
```
aantron, line 6: Please also minimize this `.dockerignore`.

example/z-docker-opam/dune-project
```diff
@@ -0,0 +1,15 @@
+(lang dune 2.0)
```
aantron, line 1: Is this file necessary for an example, given that it is a second set of redundant boilerplate?

example/z-docker-opam/deploy.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+set -e
+set -x
+
+docker-compose build
+docker-compose down
+docker-compose up --detach
```
aantron, line 8: Should this also get `LOGGING_DRIVER=journald`?

example/z-docker/Dockerfile
```diff
@@ -1,4 +1,40 @@
...
+RUN node_modules/.bin/esy build-dependencies
+
+# Build project
+ADD . .
+
+RUN node_modules/.bin/esy build
+
+FROM alpine:3.12 as run
+
+RUN apk add --update gmp libpq libev
```
ulrikstrid, line 36: With esy we can get these from the esy store to make sure they work even if we would want to use `scratch`

ulrikstrid (commented):
Just a couple of minor comments

aantron (commented):
Nice. However, I am not sure if this PR is targeted at the right example!

> Building the binaries can be done in Docker instead of copying the files over to a machine compatible with the Docker image.

The Docker example isn't doing that &mdash; it has a build completely contained on the target machine, and sends only sources. So, it should be "portable" in the sense that a macOS developer can develop locally using opam or esy, push sources, and a separate build on the remote Ubuntu machine will build the Linux container using whatever setup is there.

The Heroku example *does* involve copying binaries, because that seems like the only sensible way to deploy to Heroku at the moment. So, that example would benefit greatly from a containerized build.

An argument could be made for an example that builds Linux containers locally for the sake of reproducing a complex app setup with Docker Compose in local testing, but I think that's a slightly different kind of example.

aantron:
The Docker example is now deployed by CI (https://github.com/aantron/dream/runs/2517717269). As you can see, it consists of `rsync`ing the sources and then running the deploy script remotely, so even GitHub Actions isn't doing a build inside the CI runners. Also, the remote build is extremely fast because we get caching "for free" since there is a filesystem. That run deployed the new server in less than a minute, from push to live.

By contrast, the Heroku workflow (https://github.com/aantron/dream/runs/2511934235) actually builds the binary in CI, and sends that to Heroku. That's the part that is definitely not portable to Mac (although a local build will still work, you just can't deploy the output to Heroku, of course).

example/z-docker/Dockerfile
tmattio: Right, many of these are not required, but they are needed for typical web applications (Postgres database, npm stack, etc.). Hopefully the benefit of not having to figure out which dependencies are missing in the build context outweight the added time to install the dependencies (a few seconds). Note that this doesn't have any impact on the produced Docker image as this is only the build context: the runtime context doesn't include any of these.

example/z-docker-opam/dune-project
tmattio: I'm not sure I understand: all of the examples include a `dune-project` as they are self-contained. What did you have in mind?

example/z-docker/Dockerfile
tmattio: I'm not familiar with Esy, what would the commands to install them look like?

example/z-docker-opam/.dockerignore
tmattio: Done, I kept the default and listed the exclusion one by one, but it only contains the strict necessary now

example/z-docker-opam/Dockerfile
tmattio: Nope, removed

example/z-docker-opam/Dockerfile
tmattio: Remove!

example/z-docker/esy.json
tmattio: Same as above: removed

example/z-docker/esy.json
tmattio: Done!

example/z-docker/Dockerfile
tmattio: Done

example/z-docker/.dockerignore
tmattio: Done

example/z-docker-opam/deploy.sh
tmattio: Done

example/z-docker/Dockerfile
tmattio: Not needed, removed, thanks 🙂

example/z-docker-opam/Dockerfile
tmattio: Oh! I wasn't aware of that, might try the `debian-slim` image, unfortunately there's no version of `ocaml/opam` with it, so we either have to install opam manually, or ask for a Debian image to be pushed to the `ocaml/opam` Docker repo

example/z-docker/Dockerfile
aantron: There are more issues, like people will be looking at this example as a reference, not only using it as a starting point. They might be writing something completely different, and it will mislead them to include, for example, `postgresql-dev`, because it appears here as if it is an inherent dependency of the example.

For example, to reverse this case for a common scenario most people have probably experienced, if this was an Ubuntu container and it had a bunch of APT packages, but I was building an Alpine container, all these extra dependencies would cause me as a user to waste time looking up their Alpine equivalents &mdash even though many of the dependencies are ultimately not needed. This is very real with @ulrikstrid's suggestion to go the other way &mdash; translating Alpine packages back to Debian equivalents for `debian-slim`.

If I am in a more exotic environment, the amount of time wasted could be much greater, if I end up needlessly building from source, finding obscure versions due to some version incompatibitlies, and so on.

I don't think we can assume that these examples will be used primarily as starting points for development based on Alpine only, so I still suggest to minimize the system dependencies.

example/z-docker-opam/dune-project
aantron: This `dune-project` is used to generate an `opam` file, which is then committed. The information in the two files is redundant. All the other examples have only `(lang dune 2.0)` in their `dune-project` files.

example/z-docker/Dockerfile
aantron: esy installs them automatically. It essentially vendors a bunch of deps that would otherwise be system deps.

Compare with the instructions for this example before the PR. They give the set of necessary system deps. Everything else is installed by esy, including libev, gmp, and more.

example/z-docker-opam/Dockerfile
aantron: @ulrikstrid Can you link to an article, summary, or example of or on this?

example/z-docker-opam/.dockerignore
aantron: I recommend to add `_opam` back in, since that *will* come up in most opam projects, and the only reason it's not already an issue in the example is that it is in a monorepo :)

example/z-docker-opam/dune-project
tmattio: Isn't it because the other examples don't have an opam file?
I think generating the opam file with `dune-project` is the recommended workflow, especially for beginners, and the line `This file is generated by dune, edit dune-project instead` at the top of the opam file makes it clear that this file should not be edited.

aantron: I'm not aware of it being recommended, and it seems like more to learn for a benefit that is unclear at this point. It's an extra layer. I also don't use it myself, and no, it's not because they are *meant* to be used with esy (they are not, they build with opam as well, you just use the switch's `dream`). See this early state of the examples: https://github.com/aantron/dream/tree/65d060b9b0da77150f4f67d5252a45568226efc1/example/1-hello.

tmattio: I think having everything in `dune-project` does make it simpler for beginners: 

> it seems like more to learn

On the opposite, it removes one thing to learn, developers can just put all of their configuration in `dune-project` and they "only" have to learn one syntax instead of two.

But I don't have any strong opinion, I'll remove the `package` stanza from the `dune-project` 🙂

aantron: > On the opposite, it removes one thing to learn, developers can just put all of their configuration in `dune-project` and they "only" have to learn one syntax instead of two.

No, it's a third syntax, since you now need to understand all the extra things for that stanza, which are complex and unique to that stanza, yet the decisive file is still `opam`, lint still works on that, etc. If the `opam` client starts parsing `dune-project` files, or gets replaced by a client that does, then I will agree with this view :)

example/z-docker-opam/Dockerfile
ulrikstrid: http://crunchtools.com/comparison-linux-container-images/#Security

But mostly it's from what I've noticed when we used vurnability scanners.

tmattio:
@aantron I addressed the remaining reviews I think, apart from the `alpine` vs `debian-slim` discussion that @ulrikstrid started, but as mentionned, `ocaml/opam` does not have a `debian-slim` variant, so it would complexify the example. Not sure what's the best course of action there.

example/z-docker-opam/dune-project
tmattio: Done 🙂

example/z-docker/Dockerfile
tmattio: Done! We end up with only `libev-dev openssl-dev` for Opam and `nodejs npm perl-utils m4 linux-headers` for Esy

tmattio:
I also renamed `z-docker` to `z-docker-esy` to make it explicit that the setup uses Esy only. I can revert this part if you prefer though

example/z-docker/Dockerfile
tmattio: I think this comment is targetting the build stage then, right? The runtime dependencies are needed even with Esy (I tested and have symbol not found errors when not installing `libev`)

example/z-docker-opam/.dockerignore
tmattio: Done

example/z-docker/Dockerfile
aantron: > I think this comment is targetting the build stage then, right?

This comment was about targetting the run-time stage, not the build stage. If you looked in the original, "run-time" `Dockerfile` for this example, as I suggested, you would see the necessary system deps I was referring to:

https://github.com/aantron/dream/blob/67842e7b92f5b76058abe877cea63b89b31704af/example/z-docker/Dockerfile#L2

> The runtime dependencies are needed even with Esy (I tested and have symbol not found errors when not installing `libev`)

So, it's not surprising that testing only removing specifically libev was causing a failure &mdash; it is required even with esy, as was already known. You would have had to test the other dependencies. Your original code that was at issue in this conversation included several other libraries, like gmp and libpq, which is what prompted the discussion. However, it appears you have since slimmed them down, so this is not an issue anymore. You implicitly "tested" what me and @ulrikstrid were getting at by successfully removing all of them.

tmattio: Ok ok, thanks for the defailed explanation 🙂

ulrikstrid: I have a example somewhere on how to copy  from the esy store and reuse them in the target image

example/README.md
```diff
@@ -82,7 +82,9 @@ There are several examples showing Dream with Reason syntax.
 - [**`z-heroku`**](z-heroku#files) &nbsp;&mdash;&nbsp; to
   [Heroku](https://www.heroku.com).
 - [**`z-docker`**](z-docker#files) &nbsp;&mdash;&nbsp; on a server, using
-  Docker.
+  Docker with Esy as the package manager.
+- [**`z-docker-opam`**](z-docker#files) &nbsp;&mdash;&nbsp; on a server, using
```
aantron, line 86: The link shold point to `z-docker-opam#files`.

example/z-docker-opam/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM ocaml/opam:alpine as build
+
+# Install system dependencies
+RUN sudo apk add --update libev-dev openssl-dev
+
+WORKDIR /build
+
+# Install dependencies
+ADD hello.opam hello.opam
+RUN eval $(opam env) && opam install . --deps-only
```
aantron, line 10: AFAIK `eval $(opam env)` is not necessary if you are using the `opam` binary. It would have been necessary for `dune build` later, but `opam exec` takes care of the environment there, too.

example/z-docker/Dockerfile
```diff
@@ -1,4 +1,26 @@
...
+ADD . .
+RUN node_modules/.bin/esy build
+
+FROM alpine:3.12 as run
+
+RUN apk add --update libev
+
+COPY --from=build /build/_esy/default/build/default/app.exe /bin/app
+
+CMD /bin/app
```
aantron, line 26: Can you change this back to `ENTRYPOINT`? I vaguely understood that's the right choice from the Docker docs :)

example/z-docker/README.md
```diff
@@ -59,28 +60,28 @@ Host my-droplet
 
 SSH into your droplet:
 
-```
+```bash
```
aantron, line 63: Can you remove all these syntax settings? The GitHub bash syntax highighting unfortunately is quite poor, and it just makes a few random strings and pipe characters stick out while highlighting nothing else.

aantron (commented):
Ok, it looks good, I just left some final trivial nits :)

aantron:
> @aantron I addressed the remaining reviews I think, apart from the `alpine` vs `debian-slim` discussion that @ulrikstrid started, but as mentionned, `ocaml/opam` does not have a `debian-slim` variant, so it would complexify the example. Not sure what's the best course of action there.

I think we can merge the PR and think about this later. One option is to use Ubuntu, in hopes that it will make future `debian-slim` support from the `ocaml` images easier to transition to. However, there is not so much code here, that it will be difficult to change from Alpine to `debian-slim`.

example/z-docker-opam/Dockerfile
tmattio: You're right, will change that

example/z-docker/Dockerfile
tmattio: Done

example/README.md
tmattio: Done

example/z-docker/README.md
tmattio: Done

aantron:
Thank you! I'll set up some GitHub Actions deployment for this, and rename `z-docker`.

Issue #43, opened 2021-04-21
Support flash messages [integrate #62]


tmattio:
Flash messages are a common way to display success or error notifications to users in server-side rendered applications.
Dream already has a middleware to support sessions, which Flash messages are a specialization of, so it would be great to have a middleware and a set of API to support flash messages.

Additionnally, helpers to add flash messages in Eml templates would be really useful. For reference, here's Phoenix documentation on this: https://hexdocs.pm/phoenix/controllers.html#flash-messages. The templates look like this:

```elixir
<p class="alert alert-info" role="alert"><%= get_flash(@conn, :info) %></p>
<p class="alert alert-danger" role="alert"><%= get_flash(@conn, :error) %></p>
```

aantron:
Definitely yes! I skipped this for now because I wasn't ready to design good flash helpers myself. Any good precedents very welcome.

jsthomas:
Another precedent that might be of interest is Django's [messages framework](https://docs.djangoproject.com/en/3.2/ref/contrib/messages/). Below is my quick summary of that page.

Messages have three main parts: The body of the message, which is a string, the level (`debug`, `info`, `success`, `warning` or `error`), and one or more string-valued tags. The level is always provided as one of the tags, and custom levels are possible. Messages can be stored in different ways (in the database or with cookies, for example).

The two functions application developers are expected to use to manage messages look like:
```
add_message(request, level, message, extra_tags='', fail_silently=False)
get_messages(request)
```
the latter gives you a list of messages to iterate over. There are also convenience functions for creating messages at a particular level (e.g. `messages.info(request, "hello")`). There is functionality for disabling messages below a certain level (for example, if you don't want debug messages while running in production). The `fail_silently` option on `add_message` decides what to do if the messages framework is entirely disabled. 

By default all of the flash messages get cleared when a response is processed (so they only show up once) but the framework allows app developers to mark a message so that it won't be cleared. 

I'm interested in helping with this issue, if I can. That could mean continuing to research other precedents or putting together a prototype PR for discussion.

aantron:
Thanks @jsthomas, that saved a lot of research time for me! I would welcome a prototype PR.

aantron:
@jsthomas The example now has a [writeup](https://github.com/aantron/dream/tree/master/example/w-flash#files), is [in the playground](http://dream.as/w-flash), and I finished up the [docs](https://aantron.github.io/dream/#flash-messages) for the feature.

Thanks again for actually implementing it in #62!

[...]

PR #62, opened 2021-05-27
WIP: Add support for flash messages.

jsthomas:
This work-in-progress PR adds support for flash messages to the session middleware to address issue #43. It adds these new items:

```
val flash_messages : middleware
val put_flash : string -> string -> request -> unit
val get_flash : request -> (string * string) list
```

The function `put_flash` takes a category and message as arguments; each category/message pair is stored in a list inside a cookie. This means flash messages are limited to 4096 bytes. The stored pairs can be retrieved on the next request (and only the next request) using `get_flash`. Message categories are arbitrary strings, though one could add helper functions for common categories like `info`, `error`, etc.

A new example, `b-flash-message` illustrates how to use the two functions.

**Not yet addressed:**
- [ ] No helpers to simplify working with flash messages in eml templates.
- [ ] Documentation


aantron:
> but got stuck when I tried to generalize the `session` and `operations` record types or modify the `put` operation.

Can flash messages be stored inside fieldls of sessions? See [`Dream.put_session`](https://aantron.github.io/dream/#val-put_session).

aantron:
I don't see why flash messages need all this machinery. It seems they should be either session fields, or their own simple cookies. Do we need to ever realistically store flash messages on the server, given how ephemeral they are? If so, they can just be custom cookies. Is there a need for a database back end at all for flash messages?

jsthomas:
Thanks for your feedback. I've made some revisions to store messages in sessions instead.

aantron:
Thanks, it looks much neater now, too!

I still want to ask about sessions vs. bare cookies. With sessions, flash messages will sometimes be stored in database (depending on the session back end). Do flash messages really need that? This has potential performance implications from committing the flash message data to storage. If we directly store flash messages in a new kind of cookie (`dream.flash`?), we untangle them from sessions completely, and ensure that they always stay out of the session back end. Is that enough for flash messages? Is there any reason for them to ever end up in persistent storage server-side? (I actually don't know)

aantron:
Also, cc @tmattio.

tmattio:
Thanks for starting this @jsthomas! 🙂 

Here are my two cents after reading the PR:

- We might want to add a middleware that invalidates flash messages. With the current implementation, they are only removed when requested by the user, which allows for the weird state where old flash messages can be displayed where they have nothing to do.
- IIUC, the current PR supports setting multiple flash messages if they are of different levels. Should we support multiple same-level flash messages? Here is a relevant issue: https://github.com/phoenixframework/phoenix/issues/301

I'm also of the opinion that flash messages should be dead simple. They don't need to be signed or persisted, so they can be decoupled from sessions to keep the implementation simple.

In fact, Phoenix (sorry if I'm always putting the same framework on the table, it's the one I'm most familiar with) just provides a few functions to work with Flash:

```elixir
put_flash(conn, key, message)
get_flash(conn)
clear_flash(conn)
```

src/middleware/session.ml
```diff
@@ -355,3 +372,29 @@ let session_label request =
...
+    | Debug -> "DEBUG"
+    | Info -> "INFO"
+    | Success -> "SUCCESS"
+    | Warning -> "WARNING"
+    | Error -> "ERROR" in
+  "FLASH_MESSAGE_" ^ suffix
+
+let clear_flash level request =
+  let k = key_of_level level in
+  (fst (getter request)).remove k
```
tmattio, line 390: `remove` here does not set the cookie to a time in the past, so the flash message isn't invalidated, or am I missing something?

Also to invalidate, let's not use `Session.invalidate` which calls `Unix.timeofday`, we can use `epoch` directly and save a system call

jsthomas: Inside of `remove` we use `List.remove_assoc` to delete just the message associated with a particular level. Maybe the example application `b-flash-message` provides a good illustration? In that app, `/results` is supposed to show the flash message once and the user's text (stored in the session) every time.

>  let's not use Session.invalidate 

Agreed, that would wipe out everything in the session.

aantron:
> I personally also like APIs that offer helpers for the common levels. E.g:
> 
> ```ocaml
> val get_flash_info : request -> string option promise
> ```

If we do this, I suggest factoring out flash messages to a separate library.

jsthomas:
I spent some time researching how other frameworks store flash messages in an effort to better understand the sessions versus bare cookies issue.

Rails stores flash messages as part of the session, which means a cookie session by default (more [here](https://stackoverflow.com/questions/7331202/where-are-flash-messages-stored) and [here](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/middleware/flash.rb#L6)) but can also be a relational database. In Django, flash messages can be stored either in cookies or in the session, and the session backend can be a relational database (more [here](https://docs.djangoproject.com/en/3.2/ref/contrib/messages/#storage-backends) and [here](https://docs.djangoproject.com/en/3.2/topics/http/sessions/#using-database-backed-sessions)). Django actually goes a bit further, and has a "fallback" storage option where flash messages will be stored in a cookie if they fit and the session otherwise. Flask appears to be similar (see [here](https://flask.palletsprojects.com/en/2.0.x/patterns/flashing/) and [here](https://github.com/pallets/flask/blob/main/src/flask/helpers.py#L367)). The Flask docs have an explicit warning that flash messages that are too big may fail silently.

So there is a precedent for using session storage and/or a database to store this data, with cookie-based storage a consistent default. The downside to a cookie-only approach is that it limits flash messages to 4096 bytes. I struggled to think of a realistic use case where a flash message (or set of messages) needs to be that big. Maybe putting a stack trace or other instrumentation into a `debug` level message? Personally, I prefer see that type of data in a log.

I haven't thoroughly researched legal or security reasons that one might prefer non-cookie based flash message storage. The wikipedia article on [secure cookies](https://en.wikipedia.org/wiki/Secure_cookie#cite_note-:1-2) asserts, "even with `Secure`, sensitive information should never be stored in cookies, as they are inherently insecure and this flag can't offer real protection". This [MDN article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#cookie-related_regulations) points out that some countries have regulations about "allowing users to opt out of receiving some or all cookies". Perhaps one might want to develop an app that uses as little cookie data as possible for security or legal reasons; I haven't encountered this myself.

Overall I lean toward the bare-cookie approach because it's simple, covers most use cases, and separates concerns (no overlap with sessions).

aantron:
> Overall I lean toward the bare-cookie approach because it's simple, covers most use cases, and separates concerns (no overlap with sessions).

I agree with this. Depending on the API, it might be easy, upon future request, to offer session storage as an option. However, let's implement in cookies only to get started.

src/dream.mli
```diff
@@ -1387,7 +1387,14 @@ val session_label : request -> string
 val session_expires_at : request -> float
 (** Time at which the session will expire. *)
 
+(** {1 Flash Messages} *)
 
+val flash_messages : ?lifetime:float -> middleware
```
aantron, line 1392: Is `?litetime` necessary? We might be able to just hard-code it.

src/dream.mli
```diff
@@ -1387,7 +1387,14 @@ val session_label : request -> string
...
+type level = Debug | Info | Success | Warning | Error
```
aantron, line 1393: Can you make this a polymorphic variant, like log levels?

aantron, line 1393: Is `Success` necessary, as separate from `Info`?

src/dream.mli
```diff
@@ -1387,7 +1387,14 @@ val session_label : request -> string
...
+
+val clear_flash : level -> request -> unit
```
aantron, line 1395: Is this necessary? Shouldn't the middleware now clear the older flash messages?

src/dream.mli
```diff
@@ -1387,7 +1387,14 @@ val session_label : request -> string
...
+val put_flash : level -> string -> request -> unit
+val get_flash : level -> request -> string option
```
aantron, line 1397: Should this return a `string list`? Should it be possible to store multiple messages of the same level? Should `get_flash` return all messages of the same level or higher?

src/middleware/flash_message.ml
```diff
@@ -0,0 +1,92 @@
...
+  getter : Dream.request -> 'a;
+}
+
+
+let typed_middleware () =
+  let local = Dream.new_local ~name:"dream.flash_message" () in
+  {
+    middleware = middleware_ local;
+    getter = getter_ local;
+  }
```
aantron, line 68: You can get rid of this `typed_middleware` machinery, it is a holdover from sessions, that even sessions don't really use anymore. You can inline the getter and the middleware directly into one function, and use internal name of the local variable directly.

src/dream.mli
jsthomas: Not really, I picked these names because they were the same as the ones used by Django. I'll delete it.

src/dream.mli
jsthomas: > Should this return a string list? Should it be possible to store multiple messages of the same level?

Django allows you to store multiple messages of the same level; Phoenix appears to just let you store one (support for multiple messages was [added then removed](https://github.com/phoenixframework/phoenix/pull/304)). Looking in the [Phoenix issue](https://github.com/phoenixframework/phoenix/issues/301) linked earlier, one of the maintainers wrote:

> Going through my Ruby web apps, I almost always have a single flash message per key. The caveat of multiple is it makes the single case more verbose.

This, and the fact that multi-message support was added and then later removed makes me think the single message approach is better. But I'm happy to revise if you'd prefer the other approach.

> Should get_flash return all messages of the same level or higher?

I'm speculating, but I don't think most web developers treat flash messages' levels as totally ordered; they're more like labels for ensuring a particular message will be shown with relevant styling (green for info, red for error, etc.). It might have been better to choose some other word than "level", like "key" or "label".

src/dream.mli
jsthomas: Yes, by default the middleware will clear an old flash message on the next request. I thought `clear_flash` might be useful if one needs a way to get rid of a message created earlier in the process of handling a request (before data is written out to a cookie).

It might be better to just omit this until someone asks for it. I see that Flask does not have a function like this, for example.

src/middleware/flash_message.ml
jsthomas: I revised this part but I might have misunderstood what you had in mind regarding inlining.

src/dream.mli
jsthomas: That sounds fine to me.

src/dream.mli
aantron: > This, and the fact that multi-message support was added and then later removed makes me think the single message approach is better. But I'm happy to revise if you'd prefer the other approach.

The comments since https://github.com/phoenixframework/phoenix/pull/304#issuecomment-421169891 are about adding multiple message support, and listing workarounds.

It seems worrying that flash messages can be lost &mdash; at least for me personally, it would give me anxiety. In more "technical" terms, it makes usage of flash message non-composable, since you cannot predict, purely locally, if a message you are adding will actually be displayed, despite there not being a `clear_flash`.

Out of the frameworks I checked/confirmed so far, Phoenix and Laravel don't support multiple messages. Django, Flask, Express, and Rails do, and expect you to iterate over the messages. It seems that the most used frameworks (Rails, Express, Django) use the multiple-messages style.

I personally don't see how code can be "correct" and *not* iterate over messages, given that there are already, at the very least, multiple levels, even if with only one message each. Wouldn't you just define a sub-template for doing this iteration, and base styles on each message's "level," which you get as an "output" of `get_flash` rather than pass as an "input?"

Is there a reason to ever hold on to some flash messages for display in response to a later request? Do you want to e.g. choose only the info messages, and not display errors?

I think the base flash functionality of Dream should be "correct" and worry-free, and we can build a single-message getter on top of it later, if it is needed.

aantron: I also want to note that it is pretty easy for users to `List.filter` flash messages, or take the head, and it is easy to define the right helper based on a base functionality that doesn't ever lose messages. The converse is not true &mdash; it's not (easily) possible to define a multi-message behavior on top of a mutable, overwriting, single-message behavior, unless you wrap everything (not just the getter locally) so as to flatten lists into some kind of format inside strings, and parse them when retrieving.

aantron: If you have experience with the Phoenix style of flash messages, could you explain the advantage of

```ocaml
val get_flash : category -> request -> message (option or list)
```

over 

```ocaml
val get_flash : request -> (category * message) list
```

What benefit does Phoenix/its users gain from choosing a single category, and passing it as argument?

jsthomas: Reading this over in the last few days, I now think there are more advantages to the multi-message approach. I thought this point against the single-message approach was especially persuasive:

> it makes usage of flash message non-composable, since you cannot predict, purely locally, if a message you are adding will actually be displayed

I've revised to allow for multiple messages of the same category, and done away with fixed message categories.

example/b-flash-message/flash_message.eml.ml
```diff
@@ -0,0 +1,55 @@
...
+  <html>
+    <body>
+      Enter some text:
+      <%s! Dream.form_tag ~action:"/" request %>
+        <input name="text" autofocus>
+      </form>
+    </body>
+  </html>
+
+
```
jsthomas, line 11: Is mixing Tyxml and templates like this idiomatic? I couldn't figure out how to implement `results_page` using templates.

src/middleware/flash_message.ml
```diff
@@ -0,0 +1,56 @@
...
+
+
+let flash_messages inner_handler request =
+  let outbox = ref [] in
+  let request = Dream.with_local storage outbox request in
+  let%lwt response = inner_handler request in
+  Lwt.return(
+    let entries = List.rev !outbox in
+    let content = List.fold_right (fun (x,y) a -> `String x :: `String y :: a) entries [] in
+    let value = `List content |> Yojson.Basic.to_string in
```
jsthomas, line 24: I'm not sure if using Yojson to serialize is the right approach.

example/b-flash-message/flash_message.eml.ml
aantron: I'll edit this example to not use TyXML after the merge, and ping.

src/middleware/flash_message.ml
aantron: It is *a* right approach.

aantron:
Thanks for the PR and research!

Also thanks to @tmattio for the review.

jsthomas:
Thanks for reviewing, I learned a lot.

PR #68, opened 2021-06-12
Add template files example

mooreryan:
Hey aantron, hopefully this simple example will be useful to some people.  It shows how to put templates in their own files and how to update the dune file for it to work.

I tried to follow along with your other examples and make this one as much like those as possible.

I wasn't sure which section to put it in or which name so I went with `w-template-files`.


aantron:
Thanks! Nice write-up. I'll link this from the examples README soon-ish. I probably can't put it into the playground for now, because it only supports single-source-file examples at the moment.

aantron:
Sorry for the delay. I took a little break from Dream.

Should be all linked up now. Perhaps we will have to make this more visible over time, in case the issue keeps coming up.

Issue #75, opened 2021-06-21
discussion: implementing phoenix like websocket channels


tcoopman:
I'm looking and playing around a bit with dream and I have a hobby project that would definitely benefit from something like [channels](https://hexdocs.pm/phoenix/channels.html) in dream.

So I created this issue to start some discussion around how to implement this for dream. I haven't thought deeply about this yet, but I was wondering if you could give some input on what a good way of tackling this would be.

Questions I have at the moment:

* What would be the best way of handling runtime state
* I don't see an example of a long running websocket (the examples all send `Dream.close_websocket`) how would that work?
* I would need a reference to the `websocket` so that I can broadcast, reply to other people,... Not sure what the best way to do that either

aantron:
> I don't see an example of a long running websocket (the examples all send Dream.close_websocket) how would that work?

You can simply pass the reference that you get to the WebSocket to any other function, store it in data structures, etc. There's no reason why you must close it inside that callback.

> I would need a reference to the websocket so that I can broadcast, reply to other people,... Not sure what the best way to do that either

I think the above answers this, as well, but let me know if not.

> What would be the best way of handling runtime state

What are the options? You should be able to handle it in any way in OCaml. A simple example would be some kind of mutable map from channel names to WebSocket lists, so `(string, Dream.websocket) Hashtbl.t ref`. There might be drawbacks to that depending on what you want to implement exactly, but that's at least one way to start off with.

tcoopman:
Thanks, so, I'll create a recursive function and a mutable structure (Hashtbl for example), something like this:

```ocaml
module Channel = struct
  let joined = Hashtbl.create ...
  let join .... = update Hashtbl
end

let rec websocket_listener websocket =
  match%lwt Dream.receive websocket with
  | Some msg ->
    let key = key_from_message msg in
    let%lwt () = Channel.join ~key ~websocket in
    ws_loop websocket
  | _ ->
    Dream.close_websocket websocket
```

tested something similar to this and it seems to work.

Would this be something that would be useful to document in a simple chat example?

aantron:
Yes, if you'd like to make an example (that is still about as simple as most examples), I'd merge it :)

tcoopman:
I've been thinking a bit more about how to model this, feedback is welcome:

```ocaml
type topic =
  | Topic of string
  | WithSubtopic of (string * string)

type payload = Payload of string

type answers = answer list
and answer =
  | Reply of string
  | Broadcast of string
  | Stop of string

[...]

Issue #55, opened 2021-05-02
how to apply ocamlformat 


cdaringe:
# problem

`dune build @fmt --watch` yields

`ocamlformat: ignoring "src/bin/server/bin.eml.ml" (syntax error)`

when using `eml`

# discussion

do you use `ocamlformat`? if so, how? perhaps https://dune.readthedocs.io/en/stable/dune-files.html#dialect is relevant?

aantron:
Ah, this is indeed an oversight. I don't use OCamlFormat, as you can probably tell :)

The original intent was that as projects get more elaborate, the templates will get factored out into their own files, that are mostly HTML with only a bit of OCaml code. Those files shouldn't require formatting (at least by an OCaml formatter). They would be named like `foo.eml.html` to trigger HTML highlighting due to their mostly HTML content. I think that would bypass the OCamlFormat error.

Another option is to try refmt. Perhaps eml is similar enough to JSX. It might be awkward to set up and use for `.eml.ml` files, though.

Thanks for the link to `(dialect ...)`. It looks almost useful for this, but:

- The rule that dots aren't allowed in extensions means that `.eml.ml` files can't be distinguished from `.ml`.
- The preprocessor is required to output a binary AST, while `eml` currently outputs OCaml text.

It's likely possible to at least disable OCamlFormat on template files by either patching Dune, or by changing conventions on the Dream side.

tungd:
This is not directly answer the question but I got away with this by having templates in a separate file, using the Reason syntax and TyXML JSX support.

As a side note, I'd love it if TyXML JSX would become the "endorsed" template solution replacing the current one. I do like the built-in template solution but I'd much prefer Reason/JSX for tooling reasons.

aantron:
I think TyXML JSX (+ Melange) is basically the "endorsed" solution for Reason, and will be good for a large proportion of users. I have a TODO for making it more prominent in the README. But I don't want to offer it as the basic default, because it is Reason-only, it's an extra thing to learn, it's verbose in places, a bit non-obvious, it doesn't match JS JSX 100%, etc. So, I'd rather offer "plain, completely obvious HTML" for people that don't want to bother with JSX. But it's, of course, a great choice for people already using JSX or who absolutely need to use it in files that they need tooling support for, especially if they are also using it client-side.

@cdaringe You may want to see the [TyXML JSX example](https://github.com/aantron/dream/tree/master/example/r-tyxml#files), if you haven't already.

[...]

PR #77, opened 2021-06-22
create chat example

tcoopman:
This adds a very basic chat example. It's functional, but I'm definetely open for suggestions, changes,...

One question I have, a problem I saw, was that I sent messages to a closed websocket, which resulted in errors. I was wondering what would be the best way of handling this.
At the moment I do my best to remove the connections that will be closed, but I think race conditions are possible where you try to send something to a socket that's not there anymore.


tcoopman:
as discussed in: #75

example/z-websocket-chat/chat.eml.ml
```diff
@@ -0,0 +1,133 @@
...
+  in
+  match%lwt Dream.receive websocket with
+  | Some text ->
+      begin
+      match String.split_on_char ':' text with 
+      | [key; msg] -> 
+          Socket.add key websocket;
+          let%lwt () = Socket.send msg in
+          sockets websocket ~key:(Some key)
+      | _ -> close ()
```
tcoopman, line 114: This is not ideal, you can't have `:` in your messages at the moment. `:` is the secret close command :-)

example/z-websocket-chat/chat.eml.ml
```diff
@@ -0,0 +1,133 @@
...
+    Lwt_list.iter_p (fun w -> Dream.send w msg) websockets
+
+  let remove key =
+    Hashtbl.remove connections key 
+end
+
+let rec sockets ?(key=None) websocket =
+  let close () =
+    let () = match key with
+    | None -> ()
```
tcoopman, line 101: woops, need to close here as well. Not sure if this solution is the best though.

example/z-websocket-chat/chat.eml.ml
```diff
@@ -0,0 +1,133 @@
...
+  <div id="chat"></div>
+  <form id="form">
+      <input type="submit" value="Send" />
+      <input type="text" id="msg" size="64" autofocus />
+  </form>
+  </body>
+  </html>
+
+module Socket = struct
+  let connections = Hashtbl.create 5
```
aantron, line 85: The way this hash table is used looks wrong to me. You always send to all the sockets, regardless of key, so what's the point of the key, and why accept it from user input? What if there is a collision? I can see that you might want to retain a reference to a socket cell in a collection of sockets so that you can remove it, but for that you can probably use an internally-generated incrementing id.

If these are meant to be channels, you probably need lists of sockets (or other collection) associated with each key.

example/z-websocket-chat/chat.eml.ml
```diff
@@ -0,0 +1,133 @@
...
+
+  let add key websocket =
+    Hashtbl.replace connections key websocket
+
+  let send msg =
+    let websockets = Hashtbl.to_seq_values connections |> List.of_seq in
+    Lwt_list.iter_p (fun w -> Dream.send w msg) websockets
```
aantron, line 92: To prevent exceptions here due to a race condition, you could store a flag in each socket cell, indicating whether the socket has been closed. Alternatively, you can catch the exceptions. If the exceptions don't break the iteration, though, it's probably fine to just leave them in, since this is just an example. They probably just add noise to the log.

example/z-websocket-chat/chat.eml.ml
tcoopman: The reason for the key is to be able to clear it when the connection closes. I've started with a key from the client, because I was too fast :-), an internal reference would be better probably.

tcoopman: I've changed this to an internally generated id. I've also renamed some things to make things clearer (hopefully).
The hash table is still not that useful, just easy to work with. I could also keep every connection in a (key, websocket) list and filter these, but the hash table works.

tcoopman:
and a totally of topic question, but what editor do you use and does it manage the syntax highlighting ok? I'm using neovim and it completely breaks on the inline html (but I don't care too much about that), the issue is though that it doesn't recover and highlight the rest of the file with proper ocaml formatting :cry:

tcoopman:
Anything else you want me to change?

aantron:
Thank you! I'll make some minor edits and link the example from the example TOC.

aantron:
> but what editor do you use and does it manage the syntax highlighting ok?

I use VS Code, and it eventually recovers from mis-highlighting HTML in an `.eml.ml` file. However, in "real" projects I use `.eml.html`, where the syntax highlighting is fine (OCaml code just doesn't get highlighted, basically). See also https://github.com/aantron/dream/issues/55#issuecomment-830753551 as a related issue.

[Here's](https://github.com/aantron/dream/blob/master/example/z-playground/client/client.eml.html) an example of an `.eml.html` file on GitHub.

tcoopman:
Thanks. I'll probably also separate `eml` as much as possible from `ml`. I notice that vim recovers mostly as well, but it doesn't for this chat example.

tcoopman:
I've been thinking a bit more about how to model this, feedback is welcome:

```ocaml
type topic =
  | Topic of string
  | WithSubtopic of (string * string)

type payload = Payload of string

type answers = answer list
and answer =
  | Reply of string
  | Broadcast of string
  | Stop of string

type 'a channel =
  { handle_join : topic -> payload -> 'a * answers
  ; handle_message : 'a -> payload -> 'a * answers
  }

val channels : (string * 'a channel) list -> Dream.websocket -> unit Lwt.t
```

The idea is that you can do something like this:

```ocaml
          Dream.websocket
      @@ Socket.channels
           [ ( "public_chat"
             , { handle_join = (fun topic payload -> (initial state, [ replies ]))
               ; handle_message = (fun state payload -> (new state, [ replies ]))
               } )
           ; ( "private_chat:123"
             , { handle_join = (fun topic payload -> (initial state, [ replies ]))
               ; handle_message = (fun state payload -> (new state, [ replies ]))
               } )
           ] )
```

This is a first brain dump, so it's not complete and it has at least one big flaw at the moment. A `type 'a channel` is no good, each channel can have a different type. I'll need an other structure for that.

aantron:
Ok, the example is now [linked](https://github.com/aantron/dream/tree/de3310a8f8f239f9e8b0760977753537049ac658/example#examples), and loaded into the [playground](http://dream.as/w-chat)!

PR #80, opened 2021-06-28
Remove double slashes from sqlite3 example

thangngoc89:
Without it, it throws `Sqlite URI cannot contain user or host components`


aantron:
Thanks!

The [full example](https://github.com/aantron/dream/tree/master/example/h-sql#files) had the correct string.

PR #92, opened 2021-07-02
Add warning about common sqlite scheme mistake

### LINKED ISSUE #79: Meaningful error when using incorrect database url protocol
Update:

This error is because incorrect database protocol was used (sqlite instead of sqlite3)

——-
Instead of this 

```
web   | 29.06.21 03:06:04.924      dream.http ERROR Failure("Dream.sql_pool: cannot create pool for 'sqlite:development.sqlite': Failed to load driver for <sqlite:devel)
web   | 29.06.21 03:06:04.924      dream.http ERROR Raised at Stdlib__map.Make.find in file "map.ml", line 137, characters 10-25
web   | 29.06.21 03:06:04.924      dream.http ERROR Called from Logs.Tag.find in file "src/logs.ml", line 154, characters 14-32
```

I believe this would be the expected behavior since all of sqlite3 tools I've used do this

outkine:
Fixes #79 cc @thangngoc89

### DIFF
--- a/src/sql/sql.ml
+++ b/src/sql/sql.ml
@@ -31,8 +31,12 @@ let sql_pool ?size uri =
   begin match !pool_cell with
   | Some pool -> inner_handler (Dream.with_local pool_variable pool request)
   | None ->
+    let parsed_uri = Uri.of_string uri in
+    if Uri.scheme parsed_uri = Some "sqlite" then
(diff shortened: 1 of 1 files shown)


aantron:
Thanks!

thangngoc89:
Thank you very much ! This would help for anyone ever hitting that error

PR #100, opened 2021-07-03
Use `ptime` instead of `Unix` to compute the date (MirageOS support).

dinosaure:
This PR wants to delete the Unix dependency from the core of `dream`. It replaces `Unix.gmtime` by [ptime](https://github.com/dbuenzli/ptime) which is compatible with MirageOS (no `unix` _syscall_). It seems that both are the same so nothing seems to change.

I took the opportunity to delete `lwt.unix` as a dependency of `dream.pure` to be sure to avoid the `unix.cmxa` library at the link time.

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -71,6 +71,7 @@ depends: [
   "ssl" {>= "0.5.8"}  # Ssl.get_negotiated_alpn_protocol.
   "uri" {>= "4.2.0"}
   "yojson"  # ...
+  "ptime"  # time and date
 
   # Currently vendored.
   # "gluten"
--- a/dream.opam
+++ b/dream.opam
@@ -71,7 +71,7 @@ depends: [
   "ssl" {>= "0.5.8"}  # Ssl.get_negotiated_alpn_protocol.
   "uri" {>= "4.2.0"}
   "yojson"  # ...
-  "ptime"  # time and date
+  "ptime" {>= "0.8.1"} # time and date
 
   # Currently vendored.
   # "gluten"
--- a/src/pure/formats.ml
+++ b/src/pure/formats.ml
@@ -53,24 +53,23 @@ let to_set_cookie
     ?expires ?max_age ?domain ?path ?secure ?http_only ?same_site name value =
 
   let expires =
-    match expires with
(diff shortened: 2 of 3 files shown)


src/pure/formats.ml
```diff
@@ -53,24 +53,23 @@ let to_set_cookie
...
-        | _ -> assert false
+        match Ptime.weekday time with
+        | `Sun -> "Sun" | `Mon -> "Mon" | `Tue -> "Tue" | `Wed -> "Wed" | `Thu -> "Thu"
+        | `Fri -> "Fri" | `Sat -> "Sat"
       in
+      let ((y, m, d), ((hh, mm, ss), _tz_offset_s)) = Ptime.to_date_time time in
       let month =
-        match time.tm_mon with
+        match m with
         | 0 -> "Jan" | 1 -> "Feb" | 2 -> "Mar" | 3 -> "Apr" | 4 -> "May"
```
aantron, line 67: Per the [docs of `Unix.tm`](https://ocaml.org/api/Unix.html#TYPEtm), `tm_mon` has range 0..11. By contrast, per the docs of [`Ptime.date`](https://erratique.ch/software/ptime/doc/Ptime.html#TYPEdate), the month field has range 1..12. Could you check this pattern?

dream.opam
```diff
@@ -71,6 +71,7 @@ depends: [
   "ssl" {>= "0.5.8"}  # Ssl.get_negotiated_alpn_protocol.
   "uri" {>= "4.2.0"}
   "yojson"  # ...
+  "ptime"  # time and date
```
aantron, line 74: `Ptime.weekday` appears to [require Ptime 0.8.1](https://github.com/dbuenzli/ptime/blob/master/CHANGES.md#v081-2015-07-14-cambridge-uk).

src/pure/formats.ml
aantron: There's a good chance this is what is causing the [assertion failure in CI](https://github.com/aantron/dream/pull/100/checks#step:6:85).

aantron:
Thanks!

PR #102, opened 2021-07-04
Delete any occurrences of Unix.gettimeofday to be compatible with MirageOS

dinosaure:
This PR wants to delete the last direct dependency to `unix` into `middlewares` to be able to use them into a MirageOS application. The only _syscall_ is `gettimeofday` and it is a bit widely used.

The main change is about the `logger` which need to be initialized. In this PR, the `logger` is initialized into the `dream.ml` module. The initialization follows what we did about `mirage-crypto` and the RNG, see specially: https://github.com/mirage/mirage-crypto/pull/64

As far as I can tell, I kept the same semantic and just delayed the way to initialize the logger and still be able to plug with MirageOS/Functoria what I really want (in that case `Mirage_clock.PCLOCK`) according to my target.

Then, some parts of some middlewares are _functorized_ over `Mirage_block.PCLOCK`, some others expects a new argument `now` which is the implementation of `gettimeofday` and we partially apply them into `dream.ml`. I don't want to systematically use _functors_ so it's why some "little" functions expects only a new `now` argument. Tell me if it's fine for you.

As far as I can tell, such update does not change the interface and only internal things and how we can compose with MirageOS at the end. Tell me what do you think about all of that and how we can improve `dream` to be agnostic to the system. Again, the hard part is about the logger and I'm not really really sure how to properly handle it.

A new dependency was added: [`Mirage_clock`](https://github.com/mirage/mirage-clock) which is a simple interface. It's not an implementation.

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -72,6 +72,7 @@ depends: [
   "uri" {>= "4.2.0"}
   "yojson"  # ...
   "ptime" {>= "0.8.1"} # time and date
+  "mirage-clock"
 
   # Currently vendored.
   # "gluten"
--- a/dream.opam
+++ b/dream.opam
@@ -64,6 +64,7 @@ depends: [
   "lwt_ssl"
   "logs" {>= "0.5.0"}
   "magic-mime"
+  "mirage-clock"
   "mirage-crypto" {>= "0.8.1"}  # AES-256-GCM.
   "mirage-crypto-rng" {>= "0.8.0"}  # Signature of initialize.
   "multipart_form" {>= "0.3.0"}
@@ -72,7 +73,6 @@ depends: [
   "uri" {>= "4.2.0"}
   "yojson"  # ...
   "ptime" {>= "0.8.1"} # time and date
-  "mirage-clock"
 
   # Currently vendored.
   # "gluten"
--- a/src/dream.ml
+++ b/src/dream.ml
@@ -83,8 +83,7 @@ let log =
 
 include Dream__middleware.Tag
 
-let ( <.> ) f g = fun x -> f (g x)
-let now = Ptime.to_float_s <.> Ptime.v <.> Ptime_clock.now_d_ps
+let now () = Ptime.to_float_s (Ptime.v (Ptime_clock.now_d_ps ()))
 
 let form = form ~now
 let multipart = multipart ~now
--- a/src/dream.ml
+++ b/src/dream.ml
@@ -19,12 +19,10 @@ let next ~buffer ~close ~exn request =
 
 include Dream__middleware.Log
 include Dream__middleware.Log.Make (Ptime_clock)
-
 (* Initalize logs with the default reporter which uses [Ptime_clock], this
  * function is a part of [Dream__middleware.Log.Make], it's why it is not
  * prepended by a module name. *)
 let () = initialize ()
-
 include Dream__middleware.Echo
 
 let default_log =
--- a/src/dream.ml
+++ b/src/dream.ml
@@ -18,10 +18,11 @@ let next ~buffer ~close ~exn request =
   next ~buffer ~close ~exn request
 
 include Dream__middleware.Log
-include Dream__middleware.Echo
+include Dream__middleware.Log.Make (Ptime_clock)
 
-let logger =
-  Dream__middleware.Log.logger
+let () = initialize ()
+
+include Dream__middleware.Echo
 
 let default_log =
   Dream__middleware.Log.sub_log (Logs.Src.name Logs.default)
@@ -35,6 +36,7 @@ include Dream__middleware.Router
 include Dream__middleware.Static
 
 include Dream__middleware.Session
+include Dream__middleware.Session.Make (Ptime_clock)
 let sql_sessions = Dream__sql.Session.middleware
 
 include Dream__middleware.Flash_message
@@ -80,3 +82,12 @@ let log =
   Dream__middleware.Log.convenience_log
 
 include Dream__middleware.Tag
+
+let ( <.> ) f g = fun x -> f (g x)
+let now = Ptime.to_float_s <.> Ptime.v <.> Ptime_clock.now_d_ps
+
+let form = form ~now
+let multipart = multipart ~now
+let csrf_token = csrf_token ~now
+let verify_csrf_token = verify_csrf_token ~now
+let form_tag = form_tag ~now
--- a/src/dream.ml
+++ b/src/dream.ml
@@ -20,6 +20,9 @@ let next ~buffer ~close ~exn request =
 include Dream__middleware.Log
 include Dream__middleware.Log.Make (Ptime_clock)
 
+(* Initalize logs with the default reporter which uses [Ptime_clock], this
+ * function is a part of [Dream__middleware.Log.Make], it's why it is not
+ * prepended by a module name. *)
 let () = initialize ()
 
 include Dream__middleware.Echo
--- a/src/middleware/dune
+++ b/src/middleware/dune
@@ -5,9 +5,9 @@
   digestif
   dream.cipher
   dream.pure
+  mirage-clock
   multipart_form
   multipart_form.lwt
-  mirage-clock
   ptime
   unstrctrd
   fmt
--- a/src/middleware/dune
+++ b/src/middleware/dune
@@ -5,16 +5,16 @@
   digestif
   dream.cipher
   dream.pure
-  mirage-clock
   multipart_form
   multipart_form.lwt
-  ptime
   unstrctrd
   fmt
   fmt.tty
   logs
   lwt
   magic-mime
+  mirage-clock
+  ptime
   uri
   yojson
  )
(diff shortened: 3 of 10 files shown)


dream.opam
```diff
@@ -72,6 +72,7 @@ depends: [
   "uri" {>= "4.2.0"}
   "yojson"  # ...
   "ptime" {>= "0.8.1"} # time and date
+  "mirage-clock"
```
aantron, line 75: Not an issue for this PR, but in the future, could you add dependencies in alphabetical order?

src/dream.ml
```diff
@@ -80,3 +82,12 @@ let log =
   Dream__middleware.Log.convenience_log
 
 include Dream__middleware.Tag
+
+let ( <.> ) f g = fun x -> f (g x)
```
aantron, line 86: Nit, but this seems unnecessary (and less clear) to define just one function using this operator only twice. Unless you foresee using this a lot in this file in the future, I suggest to just define `now` below explicitly as a function of one argument, with the nested applications written out, and remove this operator.

src/middleware/dune
```diff
@@ -7,12 +7,13 @@
   dream.pure
   multipart_form
   multipart_form.lwt
+  mirage-clock
+  ptime
```
aantron, line 11: Can you also begin inserting these in alphabetical order? AFAIK the only ones out of order here are from your previous PR (but please don't reorder them during this PR, to leave the blame uncomplicated).

aantron (commented):
I did an initial review excluding the logger, continuing with that.

Please don't force-push into PRs without agreement. It complicates incremental review. We will squash the PR at the end, so a stack of weird commits is not an issue.

src/middleware/log.ml
```diff
@@ -339,90 +347,114 @@ let initialize_log
 
   enable := enable_;
   level := level_;
+  let `Initialized = initialized () in
+  ()
+
+module Make (Pclock : Mirage_clock.PCLOCK) = struct
```
aantron, line 353: Please separate this into two commits, one commit that does not indent a large chunk of code, and a second commit "Adjust whitespace after preceding commit"

i.e. wrap in the `Make` functor without indenting the parts of the body that consist of pre-existing code.

GitHub is not doing a good job of showing whitespace-only changes right now for some reason, and it makes it challenging to review this PR reading a large amount of non-changes presented as changes line-by-line.

It's also the right thing to do for future usage of blame, which I do use a lot as a form of documentation.

aantron: Even better would be not to include the "Adjust whitespace" commit, since we might tweak the PR multiple times. I'll just add  the adjustment at the end, after merging the PR.

src/middleware/dune
aantron: Also please don't resolve conversations &mdash; as the reviewer, I then have to track them down, expand them, and double check that they are actually resolved according to my feedback.

src/middleware/log.ml
dinosaure: I did something with a better diff, can I force push the first commit?

dinosaure: Or I can just push a new commit like: "Re-adjust whitespace and newline"

aantron: > Or I can just push a new commit like: "Re-adjust whitespace and newline"

This is better. If you force push the first commit, you effectively undo all the incremental work I've done, since I don't know anymore what changed. If you just add a new commit, I can choose to inspect that commit, and then inspect the net change from the start of the PR for just the files I haven't reviewed yet due to the diff (and GitHub apparently not displaying whitespace changes like it used to, at least in its dark theme).

dinosaure: Done ✅

aantron: Thanks, looks much more reviewable now!

src/middleware/session.ml
```diff
@@ -329,11 +329,15 @@ let {middleware; getter} =
...
   60. *. 60. *. 24. *. 7. *. 2.
 
-let memory_sessions ?(lifetime = two_weeks) =
-  middleware (Memory.back_end lifetime)
+module Make (Pclock : Mirage_clock.PCLOCK) = struct
+  let now () = Ptime.to_float_s (Ptime.v (Pclock.now_d_ps ()))
 
-let cookie_sessions ?(lifetime = two_weeks) =
-  middleware (Cookie.back_end lifetime)
+  let memory_sessions ?(lifetime = two_weeks) =
```
aantron, line 335: Could you do the same un-indentation here, and add a new commit with it, while I review the logger? Even though this part is small and I can review it even with the indentation mixed in, it's still better for blame to do the indentation of this chunk separately. When someone is checking the blame, they will click right past a commit that says "Adjust indentation after recent commits." Otherwise, they will face this diff and have to spend time realizing that the change here is actually much smaller than it looks in the diff, and potentially didn't even affect the parts that they are interested in. It was just indentation, but not immediately obvious.

aantron: > It was just indentation, but not immediately obvious.

Referring to the function headers with this statement.

src/middleware/log.ml
```diff
@@ -109,21 +109,20 @@ let reporter () =
 
       (* Format the current local time. For the millisecond fraction, be careful
          of rounding 999.5+ to 1000 on output. *)
-      let time =
-        let open Unix in
+      let time ~now =
```
aantron, line 112: Nit, but given that `~now` is passed to `reporter`, is there any reason to pass it to `time`, and turn `time` into a function, here?

src/dream.ml
```diff
@@ -18,10 +18,11 @@ let next ~buffer ~close ~exn request =
   next ~buffer ~close ~exn request
 
 include Dream__middleware.Log
-include Dream__middleware.Echo
+include Dream__middleware.Log.Make (Ptime_clock)
+
+let () = initialize ()
```
aantron, line 23: This should probably be qualified with a module path or at least commented, because right now, it is unclear that this function has something to do with logging, or where to look it up.

src/middleware/log.ml
```diff
@@ -339,15 +348,39 @@ let initialize_log
...
+  let initialize () =
+    if !set then Logs.debug (fun m -> m "Dream__log.initialize has already been called, \
+                                         ignoring this call.")
+    else begin
+      ( try let `Initialized = initialized () in
+            Format.eprintf "Dream__log.initialize has already been set, check that \
+                            this call is intentional" ;
+        with Logs_are_not_initialized -> () ) ;
+      set := true ;
+      _initialized := Some initializer_
```
aantron, line 377: Can you comment (in the PR) on the need for this two-step initializer (a function, followed by the old lazy initializer)? I suspect it's a way to avoid functorizing too much of the code, i.e. you are trying to inject the initializer from a functor into non-functorized code. But I'd just like to confirm it with you.

dinosaure: So I tried to keep the same semantic as before (with `enable` and other values with side-effect and the lazy way to prepare the `reporter`) and put a new indirection. So I kept the `lazy` things as before but we must (see `functoria`/`mirage`) have a _functor entry-point_.

> you are trying to inject the initializer from a functor into non-functorized code.

So yes, this is the final goal to let `functoria` then to apply `dream` with a specific implementation of the posix-clock **and** to not be invasive about what we already have :+1:. But I agree that it's a bit complex to understand especially when the only current case is to apply with `Ptime_clock` - but it's required for MirageOS.

src/dream.ml
dinosaure: I added a comment when the function comes from the result of the functor above. We can rename it if you want.

src/middleware/log.ml
dinosaure: No reason, however it should be better to add a unit argument may be to show the **real** time when we make the log's line (and do not pre-compute `time`)? It's minor and it can be a part of another PR (unrelated to MirageOS).

aantron: > No reason, however it should be better to add a unit argument may be to show the **real** time when we make the log's line (and do not pre-compute `time`)? It's minor and it can be a part of another PR (unrelated to MirageOS).

I don't understand the intent behind this comment. On the one hand, getting the actual time is already suspended because `now` is a function of `unit`, and we can just move the application of `now` somewhere else.

On the other hand, this comment sounds like you are suggesting we should delay evaluating the time. But don't we want to advance evaluating the time instead, to the soonest point when the log was actually executed? It's already quite late as it is, in the reporter. I think that's probably fine, it's okay to accept a small amount of imprecision.

dinosaure: > On the other hand, this comment sounds like you are suggesting we should delay evaluating the time. But don't we want to advance evaluating the time instead, to the soonest point when the log was actually executed? It's already quite late as it is, in the reporter. I think that's probably fine, it's okay to accept a small amount of imprecision.

Yes, I'm not really clear with my english, sorry. But this what I tried to say. So you can do what you want, just to notice you this specific point 🙂 .

aantron:
Thank you!

PR #103, opened 2021-07-05
Fix sandbox issue on jsoo example

### LINKED ISSUE #69: (newb) `npx esy` fails but separate dune instructions work [waiting on #63]
(I'm not being self-deprecating, I really am a newbie at OCaml, `dune` and `esy`, so it is almost certainly something I'm doing!)

Hi - I'm on macOS and trying out https://github.com/aantron/dream/tree/master/example/w-fullstack-jsoo. Following the `esy` route produces:
```
➜  w-fullstack-jsoo git:(master) ✗ npx esy
info esy 0.6.10 (using esy.json)
cp: static/client.js: Operation not permitted
error: command failed: 'cp' '_build/default/client/client.bc.js' 'static/client.js' (exited with 1)
esy-build-package: exiting with errors above...
error: build failed with exit code: 1
 
esy: exiting due to errors above
```

However, if I do the listed `dune` steps then after following the `dune`/`opam` prompted instructions to install missing bits then it all works. If I run `npx esy` again I get the above error.

I also got a similar experience with the other full stack tutorials.

I haven't the first clue as to _what_ is causing the problem so I thought I would start here :-).

ulrikstrid:
Allow creating folders in root of project by setting `buildsInSource` to `unsafe`

Fixes #69

### DIFF
--- a/example/w-fullstack-jsoo/esy.json
+++ b/example/w-fullstack-jsoo/esy.json
@@ -14,7 +14,7 @@
     "@opam/conf-libev": "esy-packages/libev:package.json#0b5eb6685b688649045aceac55dc559f6f21b829"
   },
   "esy": {
-    "buildsInSource": "_build",
+    "buildsInSource": "unsafe",
     "build": [
       "dune build --root . client/client.bc.js",
       "mkdir -p static",


ulrikstrid:
Another fix might be to create the `static` directory inside of `_build`

aantron:
Thanks!

@ulrikstrid, would you mind checking whether the other fullstack examples work on your Mac? [w-fullstack-rescript](https://github.com/aantron/dream/tree/master/example/w-fullstack-rescript#files) and [w-fullstack-melange](https://github.com/aantron/dream/tree/master/example/r-fullstack-melange#files). They work on my machine, but so did w-fullstack-jsoo.

ulrikstrid:
Sure, I can try that

PR #108, opened 2021-07-05
Delete usage of Lwt_io to be compatible with MirageOS and use `prerr_string` instead

dinosaure:
Not sure about this little PR but it wants to replace `Lwt_io.(write stderr)` by `prerr_string`. On the [`ocaml-freestanding`](https://github.com/mirage/ocaml-freestanding)/[`Solo5`](https://github.com/Solo5/solo5), we still write to the serial port 1, I believe that nothing change so.

### DIFF
--- a/src/middleware/log.ml
+++ b/src/middleware/log.ml
@@ -92,7 +92,7 @@ let reporter ~now () =
       (* Write the message. *)
       Lwt.async begin fun () ->
         Lwt.finalize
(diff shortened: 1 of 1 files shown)


aantron:
Thanks. We can then probably delete the whole surrounding `Lwt.async` call, but I'll do that separately. No need to block the PR for that.

Issue #105, opened 2021-07-05
The Playground needs better styling [CSS]


aantron:

tcoopman:
What do you want to change?

thangngoc89:
What are you thinking about? I know some CSS

aantron:
In general, the CSS is a bit ugly and not polished. I think the overall layout is pretty much right (including collapsing down on thin viewscreens). But it just doesn't have that polished, legible feel of nice, thought-out spacing, etc. In addition, if colors could be made better, it would be nice to see how :)

thangngoc89:
I didn't notice until you said it. The padding and spacing are a little bit off. I don't write pure css these days but let's me see what I can do.

aantron:
@thangngoc89 in https://github.com/aantron/dream/pull/109#issuecomment-874299534:

> I think we would need to design this in a design program to see how things go. It's easier to move things around by drag and drop then write css code.

Probably also a good idea :)

aantron:
A relatively good way to develop the playground is to save the HTML and/or JS locally, and then run it. Maybe I should put some effort into making the playground more easily deployable/runnable locally.

aantron:
Closing this for now, after several improvements by @thangngoc89 and @artileda.

More improvements are still welcome, however, if you spot anything. Or just think the Playground looks ugly :)

PR #109, opened 2021-07-05
Playground CSS improvements

thangngoc89:
This PR addresses #105 in attempt to make the playground looks nicer.
Things done:
- Vertically aligned headers and items inside it
- Change Run button color + add border radius

All numbers are based on scale of 4px.

A picture is better than a thousand words so here we go:

image

### DIFF
--- a/example/z-playground/client/playground.css
+++ b/example/z-playground/client/playground.css
@@ -80,14 +80,12 @@ h1 {
   background-color: #3730A3;
 }
 
-/* These code doesn't seem to be needed anymore */
-/* 
 #editor header > a {
-  float: right;
   color: inherit;
   text-decoration: none;
-  margin-right: 2em;
-  margin-top: 28px;
+  flex: 1;
+  text-align: right;
+  margin-right: 24px;
 }
 
 @media (max-width: 550px) {
@@ -95,7 +93,7 @@ h1 {
     display: none;
   }
 }
-*/
+
 /* width */
 ::-webkit-scrollbar {
   width: 10px;
--- a/example/z-playground/client/playground.css
+++ b/example/z-playground/client/playground.css
@@ -39,14 +39,17 @@ body {
 }
 
 
-
+header {
+  height: 64px;
+  display: flex;
+  align-items: center
+}
 /* Editor */
 
 h1 {
   margin: 0;
-  height: 21px;
-  padding: 21px 0 20px 32px;
   display: inline-block;
+  margin-left: 24px;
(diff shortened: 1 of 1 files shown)


example/z-playground/client/playground.css
```diff
@@ -61,20 +64,24 @@ h1 {
...
 }
 
-button:hover {
+#editor button:hover {
   cursor: pointer;
-  background-color: rgba(255, 255, 255, 0.3);
+  background-color: #3730A3;
 }
 
+/* These code doesn't seem to be needed anymore */
```
thangngoc89, line 83: @aantron You probably want to test this locally to confirm it since I can't build the playground myself. But I'm 99% sure that this could be deleted

example/z-playground/client/playground.css
```diff
@@ -39,14 +39,17 @@ body {
 }
 
 
-
+header {
+  height: 64px;
```
thangngoc89, line 43: @aantron Since I'm using browser's save as to adjust the styling, log doesn't show up for me so I'm not sure if this would affected the log and iframe or not. Please check this locally

example/z-playground/client/playground.css
aantron: Some modified form of these are needed, I think, for this link, which now looks like:

Снимок

It only appears in "programs" that correspond to actual documented examples, such as http://dream.as/1-hello. It's live now in this purple unstyled form :)

example/z-playground/client/playground.css
aantron: It looks correct!

aantron:
Thanks! Thsi branch is live now. Except for the thing with the link, it's ready to merge.

thangngoc89:
@aantron I updated that. It looks just like before :

image

aantron:
It looks misaligned now that everything is aligned :) Or would you prefer to keep it that way in this PR?

aantron:
Alright, so it's actually aligned with the button, but because of the button colors, the eye wants the baseline of the link to be aligned with the baseline of the title. That can just be "fixed" later or ignored. Thanks!

thangngoc89:
Run and `Docs for this example` are actually aligned but it looks misaligned because of the color illusion. I think we would need to design this in a design program to see how things go. It's easier to move things around by drag and drop then write css code.

image

aantron:
It's now live and already looks considerably better. I guess even the button alone is like 10% of the UI or so.

Issue #99, opened 2021-07-03
Add an example with PostgreSQL and Docker Compose


aantron:
...to serve as an easy starting point for people that want that kind of setup.

We already [have an example](https://github.com/aantron/dream/blob/b8a46f4649d9ca296d89b49acd55aed8a7e29a41/example/r-template/dune#L5-L8) that just uses esy, Dream, and Docker Compose. We just need to:

- [ ] Copy it and add a Postgres container to it.
- [ ] Connect to the database from the Dream code. See the [SQL example](https://github.com/aantron/dream/tree/b8a46f4649d9ca296d89b49acd55aed8a7e29a41/example/h-sql#files). We need to do something similar, but use a Postgres connection string rather than SQLite.
- [ ] Remove deployment boilerplate, as it won't be a deployment example. We can link back to the Docker deployment example for that.

PR #112, opened 2021-07-05
Add an example with PostgreSQL and Docker Compose.

jsthomas:
This change addresses #99. It introduces an example illustrating how to use Docker Compose to run a Dream container and a PostgreSQL container. The example app server is essentially the same as an earlier [example](https://github.com/aantron/dream/tree/master/example/h-sql#files) with small changes to use PostgreSQL instead of SQLite.


aantron:
@jsthomas, looks great!

To get the CI to pass, could you, before this line

https://github.com/aantron/dream/blob/dd2b8dcdf458d0b95f91e0ae058bdfd71e28672a/.github/workflows/test.yml#L31

add `- run: opam depext --yes conf-postgresql`

and before this line

https://github.com/aantron/dream/blob/dd2b8dcdf458d0b95f91e0ae058bdfd71e28672a/dream.opam#L99

add `"caqti-driver-postgresql" {with-test}`

jsthomas:
Ok, I believe CI should be all set now. Thanks for helping me fix that.

aantron:
Thanks for the PR and the fix!

PR #113, opened 2021-07-06
Log: avoid Lwt.async

hannesm:
Now that the reporter does not use Lwt anymore, remove the Lwt.async call.


aantron:
As noted in #108. Thanks!

PR #114, opened 2021-07-06
dream.pure is MirageOS compatible

dinosaure:
This is the last occurrence which brings the `unix` dependency into `dream` - I hope.


aantron:
Thanks!

PR #107, opened 2021-07-05
Use bigstringaf instead of Lwt_bytes to remove unix dependencies

dinosaure:
`Lwt_bytes` comes with some _syscalls_ such as `read`/`write` on Unix `file-descriptor`. This PR is a simple replacement of `Lwt_bytes` by [`bigstringaf`](https://github.com/inhabitedtype/bigstringaf) which is compatible with MirageOS.

PR #115, opened 2021-07-07
`dream.middleware` and `dream.cipher` is MirageOS compatible

dinosaure:
This patch is small but a bit complex. `mirage-crypto-rng.lwt` is not compatible with MirageOS because it plays with `Lwt_engine` to add a hook to continuously feed the random number generator - where MirageOS has its own scheduler. So we need to delete it from `dream.cipher` to be MirageOS compatible.

However, at this step, we have multiple choices:
- _functorize_ as we usually do with [`Mirage_random.S`](https://github.com/mirage/mirage-random)
- Strictly use `mirage-crypto-rng` and initialize it into the main module `dream.ml` (as we did about logs - see #102)

I chosen the second solution but from the abstraction perspective, we have a strong link between `dream` and `mirage-crypto-rng` (even for MirageOS). It implies that the end-user must trust on `mirage-crypto-rng` and he is not able to chose something else then - even for MirageOS.

This is an hard issue and well-know in MirageOS, so I just would like to have the opinion of @hannesm on that and, of course, your opinion @aantron.

The second commit is an implication of the first one. Due to the deletion of `mirage-crypto-rng.lwt` from `dream.cipher` and with my others patches (see #100, #102, #107, #108 and #114), `Lwt_engine`/`Lwt_unix`/`Lwt_io` are not available anymore via subsequent dependencies. So, with the first patch, `static.ml` does not compile (which requires `Lwt_io`). So I decided to, **internally**, make a sub-package `dream.middleware.static` (which brings `lwt.unix`) and includes it into the main `dream.ml` module.

Such design gives me the opportunity to not use `static` middleware into my MirageOS (which does not have a file-system). Of course, at the end, nothing change on the Dream interface.


hannesm:
My point of view: Indeed for cryptographic keys you need a strong (cryptographically secure) random number generator. This also needs to be initialized (and best periodically fed entropy into). Using Mirage_crypto_rng is a good choice. The `generate` function allows an optional custom generator `g` being passed -- in the case you want something different from the standard Fortuna one (this `g` could be exposed if there are users who wish to use their own).

Initialization, as done here, is fine. The `Mirage_crypto_rng_lwt.initialize` adds a periodic task (`Lwt.async`) and adds a callback to Lwt_main.Enter_iter_hook -- both are good to be done at the top level (they won't actually do something until Lwt_main.run is executed).

src/dream.ml
```diff
@@ -64,6 +64,8 @@ include Dream__middleware.Site_prefix
 let error_template =
   Dream__http.Error_handler.customize
 
+let () = Mirage_crypto_rng_lwt.initialize ()
```
aantron, line 67: The initialization should remain lazy. Part of Dream's design (at least outside Mirage) is that it should do as little invasive initialization as possible, unless it is actually called into. For example, Dream might be linked into some binary that has multiple sub-commands, and one command that makes the binary serve a web API for the user to locally connect a browser to. When the program is not doing Dream-related things, Dream shouldn't be feeding entropy into an RNG.

It looks like the approach from the log PR would work here &mdash; create a `ref` of what `lazy` will be forced by `random_buffer`, and set it here for the Unix/Lwt version of Dream.

src/middleware/dune
```diff
@@ -21,6 +24,19 @@
...
+ (name dream__static)
+ (modules static)
+ (libraries
+  digestif
+  dream.pure
+  lwt.unix
+  magic-mime
+ )
+ (preprocess (pps lwt_ppx))
+ (instrumentation (backend bisect_ppx)))
```
aantron, line 38: Rather than doing it this way, I recommend to create `src/unix/`, inside it `dream.unix`, and move `Dream.static` middleware there. Then, we can put any other Unix-only things that arise over time into that directory &mdash; should be more future-proof.

Is there any analogue to `Dream.static` that would make sense for Mirage?

src/dream.ml
dinosaure: I tried to follow what I did about logs for the random initialization, hope that it's good for you. May be we should add some comments to explain the pattern and why we need such indirection (reference and lazy).

src/cipher/dune
```diff
@@ -3,6 +3,7 @@
  (name dream__cipher)
  (libraries
   cstruct
+  logs
```
aantron, line 6: Can you simplify the pattern? We can essentially guarantee during "composition" that the `ref` will be initialized. I don't think it's worth picking up a dependency on `logs` in `dream.cipher` for the sake of somewhat elaborate messages about initialization problems :) I suggest deleting all natural-language messages and just doing a simple `ref`, dereferencing of it, and setting of it in `dream.ml`, only.

src/cipher/random.ml
```diff
@@ -8,7 +8,46 @@
 (* TODO LATER Is there something with lighter dependencies? Although perhaps
    these are not so bad... *)
 
+exception Entropy_is_not_initialized
```
aantron, line 11: We don't need this exception. We can just use `failwith` in the `None` case, or even `assert false`. It should be unreachable because of program composition, so we just need some fatal error that will either point people to the source line, or include a message about what it is, if we ever mess up.

src/middleware/dune
dinosaure: > Is there any analogue to Dream.static that would make sense for Mirage?

MirageOS has nothing at the beginning and, according to your application, you can "plug" a file-system which should respect a certain interface. Currently, we have [`Mirage_kv.S`](https://github.com/mirage/mirage-kv) as a common interface to get some values from a key and [`ocaml-crunch`](https://github.com/mirage/ocaml-crunch) or [`docteur`](https://github.com/dinosaure/docteur) as a possible implementation of this specific interface.

It's what we can propose to the MirageOS user as a _read-only_ "file-system". However, such choice is owned by the end-user and we should keep the ability to make an operating system without a file-system and with `dream`. My idea is to let the user to use a `static` middleware from a `Mirage_kv.S` module if he wants - but `dream.mirage` should not integrate it as a _standalone_ middleware.

src/cipher/random.ml
```diff
@@ -8,7 +8,46 @@
...
+          "Dream__cipher.Random.initialize has already been set, check that this call \
+          is intentional";
+        with
+          Entropy_is_not_initialized -> ());
+    set := true ;
+    _initialized := Some (Lazy.from_fun f)
+  end
+
 let random_buffer n =
+  let `Initialized = initialized () in
```
aantron, line 50: AFAICT there is no need for this variant, either. In summary, all these extra features, to my understanding, tend to only obfuscate what is going on here, which is we are putting some `Lazy.t` into a `ref`.

src/dream.ml
aantron: Yes, I was already considering comments, too. However, seeing the version of the logs initialization in *this* context, it looks substantially more complicated than it needs to be. So, I suggested some simplifications, and we can write comments for the simplified version (if you agree).

src/middleware/dune
aantron: Thanks for telling me about that, it makes sense, and I agree!

src/cipher/random.ml
```diff
@@ -8,40 +8,22 @@
...
-   \n  If you are using Lwt/Unix, execute `Dream.random_initialize ()`
-   \n  If you are using MirageOS, use the Dream device in config.ml
-   \n"
-
-let () = Printexc.register_printer @@ function
-  | Entropy_is_not_initialized ->
-    Some ("The entropy is not yet initialized. " ^ setup_entropy)
-  | _ -> None
-
 let set = ref false
```
hannesm, line 11: this `set` can be removed.

src/cipher/random.ml
```diff
@@ -8,40 +8,22 @@
...
-        let `Initialized = initialized () in
-        Format.eprintf
-          "Dream__cipher.Random.initialize has already been set, check that this call \
-          is intentional";
+        let `Initialized = initialized () in ()
         with
-          Entropy_is_not_initialized -> ());
+          Failure _ -> ());
     set := true ;
     _initialized := Some (Lazy.from_fun f)
```
hannesm: this function has rather complex semantics, it should just "_initialized := Some (Lazy.from_fun f)".

as it is now, it first calls to `initialized` (why?) - which always fails! and then sets two reference cells.

it reminds me a lot on the initialization in Mirage_crypto_rng -- which is mainly there to avoid pure unix and lwt mixing (+ informative messages if these are mixed in a binary).

hannesm: (the semantics is tricky since the case where `_initialized` is already `Some _`, you'll force the value here.

src/cipher/random.ml
```diff
@@ -8,28 +8,22 @@
...
-        let `Initialized = initialized () in ()
-        with
-          Failure _ -> ());
-    set := true ;
-    _initialized := Some (Lazy.from_fun f)
-  end
+  ( try
+      initialized () ;
+    with
+      Failure _ -> ());
```
hannesm: suggestion: remove this `( try .. );` block.

dinosaure: Done :+1:.

src/dream.ml
hannesm: > When the program is not doing Dream-related things, Dream shouldn't be feeding entropy into an RNG.

TBH I don't really think it is worth to complicate the initialization for this niche case. If you can measure the RNG feeding, that should be very concerning for the general approach. I wouldn't bother with the Lazy, and just call `Mirage_crypto_rng_lwt.initialize ()` in the top-level of Dream.

aantron: I would agree in principle, but just a Lazy is not that complicated, and the code in this PR could be simplified even more while keeping it (though it is already quite simple &mdash; thanks @dinosaure for making the edits!). I still would like Dream to be non-invasive whenever it is at all feasible. Apart from other considerations, it's possible (even if *extremely* unlikely) that in another part of the app, it might want to do some kind of different initialization of `Mirage_crypto_rng` for some reason. I just don't want Dream to have any unnecessary effects or co-effects when not being used.

aantron:
Thanks @dinosaure for PR and @hannesm for review!

PR #117, opened 2021-07-08
Fix the compilation of dream on OCaml 4.11

dinosaure:
A simple fix when OCaml 4.11 complains about weak type. Sorry for the noise.


aantron:
Thanks! I guess it's time to expand the build matrix in Actions.

PR #126, opened 2021-07-10
Goodbye float, flexbox all the way

thangngoc89:
As I've said in https://github.com/aantron/dream/pull/124#issuecomment-877633060

This PR removes all the fragile pixel calculations and use flexbox (shrink-grow) to control the overall layout.

I also took the opportunity to move <kbd>Change View</kbd> to the right of <kbd>Run</kbd> button because I think <kbd>Change View</kbd> is rarely used comparing to <kbd>Run</kbd>


aantron:
Thanks!

PR #127, opened 2021-07-10
Prevent Codemirror to grow

thangngoc89:
When there're too many lines, the editor would grow to fit the whole content.
The usual fix would be defining a maximum height on parent container of `.CodeMirror` 
and `position: relative`.

Since parent container is this case (`#textarea`) grow with its container (`#editor`), 
maximum height can't be set without resorting to manual `calc` .

This PR implement a hacks documented here: https://discuss.codemirror.net/t/size-inside-flexbox/1359/5

You can test this by making the editor has more lines than its parent container.
For example: http://dream.as/k-websocket

### DIFF
--- a/example/z-playground/client/playground.css
+++ b/example/z-playground/client/playground.css
@@ -50,9 +50,15 @@ body.full-editor {
 
 #textarea {
   flex: 1 0 auto;
+  position: relative;
 }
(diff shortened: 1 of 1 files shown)


aantron:
Deployed and confirmed, thanks!

PR #130, opened 2021-07-14
Make `Dream.redirect ?status` typesafe

### LINKED ISSUE #129: Dream.redirect should limit to 30x status code only and return empty body
Related code: https://github.com/aantron/dream/blob/65738dfbf1b85fba6e8612852d0e70b60a474d15/src/pure/inmost.ml#L473-L481

Limit `status` to `Dream.redirection` type

thangngoc89:
This fixes #129 
As discussed on Discord, if user passes `?code` directly, Dream would respect that without any checks. 

If `code != 30x` then browsers don't redirect the new `Location`

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -451,7 +451,7 @@ val json :
     {!Dream.application_json}. *)
 
 val redirect :
-  ?status:(redirection :> status) ->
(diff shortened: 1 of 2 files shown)


aantron:
Thanks!

Issue #134, opened 2021-07-18
Tidy up the flash message cookies


aantron:
The [`Dream.flash_messages`](https://aantron.github.io/dream/#flash-messages) middleware currently always sets a `dream.flash` cookie &mdash; it's just empty if there are no flash messages. The same empty, unconditional `dream.flash` cookie is also used to effectively "delete" flash messages after they have been "received" by the next request, by overwriting them with the same empty message list.

This can be observed by running the [`w-flash` example](https://github.com/aantron/dream/tree/master/example/w-flash#files) and looking in the network tab, where a `dream.flash` cookie will appear on the very first request &mdash; even though there are no flash messages set yet.

What should happen instead is:

- If there were no flash messages, and none are added during a particular request, no cookie should be set.
- If there were flash messages, and none are added during a request, the cookie should be deleted by setting [`~expires:0.`](https://aantron.github.io/dream/#val-set_cookie) on the returned cookie.

The "lightly offending" code is here:

https://github.com/aantron/dream/blob/5f6b7f508193a1423acd91ee2c699857ca45fa29/src/middleware/flash.ml#L24-L33

PR #137, opened 2021-07-19
Tidy up flash message cookies.

jsthomas:
This change addresses #134. In the first flash message implementation, the middleware would create a cookie regardless of whether any messages were actually created. A better approach is to only create a cookie when there is a flash message to display on the next request and to expire an existing cookie if there are no new messages to show.


aantron:
Thank you!

Issue #98, opened 2021-07-03
Add an example with nginx or other reverse proxy


aantron:
While Dream is designed to be able to run on the modern Web without a proxy (e.g., it has direct HTTP/2, HTTPS support, etc.), in practice, depending on deployment, Dream will still often run behind a proxy. So, we want to make sure Dream and proxies interact well. At least a basic nginx example will help people set up quickly, and will give us an easy way to work out proxy header-reading middleware. See #10. We should probably document whatever comes out of #10 in the nginx example.

Respecting proxy headers such as `X-Forwarded-Proto` (#10) is actually a security issue (in terms of defense in depth, rather than actual vulnerabilities), since we need to know if nginx served as an HTTPS endpoint (even if Dream gets a non-HTTPS connection from nginx) to, for instance, have Dream automatically choose more secure cookie settings that require knowing that we are running under HTTPS.

cc @thangngoc89

thangngoc89:
I would really love to help but I don’t know nginx actually. It’s auto configured for me by dokku 😭

aantron:
Done by @jsthomas in #140.

This makes it much easier to address #10 for at least nginx.

Issue #10, opened 2021-04-08
Add trust_proxy_headers middleware


aantron:
...which would allow opting in to, one by one, trusting:

- [ ] [`X-Forwarded-Host`, `Forwarded`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded)
- [ ] [`X-Forwarded-For`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)
- [ ] [`X-Forwarded-Proto:`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Proto)
- [ ] Some request ID header, to use an extrenal ID.

PR #140, opened 2021-07-21
Add an example showing how to use nginx as a reverse proxy.

jsthomas:
This change addresses Issue #98 by adding a new example showing how to use nginx as a reverse proxy together with Dream. To simplify setup, both nginx and the example server run in docker containers managed by docker compose. In the example, we show how to use nginx to handle requests for static assets so that that traffic never reaches the application server.

Issue #98 mentions security concerns around respecting proxy headers. The example introduced here doesn't cover those concerns so it won't allow #98 to be closed but may serve as a useful starting point for running those kinds of tests in the future.


aantron:
Thanks! I just started an `nginx` branch locally but didn't actually get anything done yet, because of preemption by other tasks. I will review this shortly. It's fine to consider having an nginx example as closing #98 &mdash; #10 is the issue about trusting/reading the proxy's headers, and solving #10 at least for nginx will be made considerably easier by having this example!

example/w-nginx/README.md
```diff
@@ -0,0 +1,91 @@
...
+  ```
+
+<br>
+
+**See also:**
+
+- [**`z-docker-esy`**](../z-docker-esy#files) deploys to Digital Ocean with
+  Docker Compose and esy, including Docker installation instructions.
+- [**`z-docker-opam`**](../z-docker-opam#files) deploys with Docker Compose and
+  opam.
```
aantron, line 87: Another good link here could be [`f-static`](https://github.com/aantron/dream/tree/master/example/f-static#files), in case someone wants to see how to serve static files without nginx.

example/w-nginx/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+This will build and start the [two
+containers](https://github.com/aantron/dream/blob/master/example/w-nginx/docker-compose.yml),
+one for nginx and one for the application server. The first build of
+will take several minutes. Later builds will be faster, due to
+caching.
+
+Visit [`http://localhost:8080`](http://localhost:8080) to reach the
+application server directly. Notice how an image (the static asset)
+fails to load and we just see the alt text, "The OCaml logo."
```
aantron, line 53: What would the user do to prevent direct public access to the backend server?

example/w-nginx/README.md
```diff
@@ -0,0 +1,91 @@
...
+
+
+let () =
+  Dream.run ~interface:"0.0.0.0"
+  @@ Dream.logger
+  @@ Dream.router [
+    Dream.get "/" (fun _request -> Dream.html body)
+  ]
+  @@ Dream.not_found
+```
```
aantron, line 36: I recommend also linking and/or showing parts of the nginx configuration, and linking to nginx docs rather than its home page.

example/w-nginx/default.conf
```diff
@@ -0,0 +1,30 @@
...
+    #
+    # Note: If you wish to provide server access behind another route,
+    # this can be accomplished by changing location above (say, to
+    # /app) and then adding a rewrite rule like this:
+    #
+    # rewrite ^/app/(.*) /$1 break;
+    #
+    # Note: In the proxy_pass command below, the "web" in
+    # http://web:8080 need to match the docker container name and port
+    # exposed in docker-compose.yml.
```
aantron, line 26: IMO this comment should be worked into the example text somehow, where it is much more visible, and that will also be a good place to link to the `.conf` file and show snippets of it. Based on the content, you will also have good reason to link to the `docker-compose.yml` file. Let me know if you want me to merge the PR "early" so you can easily get accurate permalinks to the files for a second round of editing. I usually do examples that way myself: I write and push something approximate, then go back and get the right permalinks for everything from the browser, and hyperlink it all together.

example/w-nginx/nginx.conf
```diff
@@ -0,0 +1,34 @@
...
+    include       /etc/nginx/mime.types;
+    default_type  application/octet-stream;
+
+    # Expose more information from the nginx logs.
+    log_format  main  '$remote_addr - $remote_user [$time_local] "$request" '
+                      '$status $body_bytes_sent "$http_referer" '
+                      '"$http_user_agent" "$http_x_forwarded_for"';
+
+    access_log  /dev/stdout;
+    error_log   /dev/stderr;
```
aantron, line 24: Are these settings redundant with the ones in `default.conf`? Also, if we have total control of the server and are only serving one site, would it be better to merge the configs? Or is that considered a bad practice for some reason?

jsthomas: The main difference from the default settings are lines 19-22 (changing the log format) and the include statement on line 33. The logs on the default nginx container are so terse that I found it hard to see what's going on.

I'm not an nginx expert by any means, but typically when I had to manage several different resources behind the reverse proxy, I did prefer to break those out into smaller files. Looking at [this discussion](https://serverfault.com/questions/527630/difference-in-sites-available-vs-sites-enabled-vs-conf-d-directories-nginx#527644), specifically the comments from `adamczi`, it looks like people do fold their site configuration into `nginx.conf` when they're deploying nginx in a container.

I think it depends a bit on whether the example is supposed to show how to use nginx in some generality (in which case, multiple configs are a bit more idiomatic), or if it instead should illustrate setting up a _containerized_ reverse proxy (in which case one config is better).

aantron: I say to show a containerized version because...

- It is containerized in this example.
- Containerized is a pretty common way to do it nowadays.

There is a third reason to *strongly* prefer configuration in one file:

- The examples must be as concise as possible, subject to still being correct and informative. If two files are not necessary, and not even the norm for this particular kind of setup, there is no reason to pay the attention cost of the reader figuring out two `.conf`  files, trying to understand the relationship between them, etc.

example/README.md
```diff
@@ -117,6 +117,8 @@ if something is missing!
   watcher using fswatch.
 - [**`w-live-reload`**](w-live-reload#files) &nbsp;&mdash;&nbsp; a simple
   live-reloading setup.
+- [**`w-nginx`**](w-nginx#files) &nbsp;&mdash;&nbsp; a simple example
+  illustrating how to use nginx as a reverse proxy.
```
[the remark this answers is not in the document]
jsthomas: Good point, I will make this more concise.

example/w-nginx/nginx.conf
jsthomas: That makes sense. I've combined the two configs and made the combined config more concise.

example/w-nginx/README.md
jsthomas: There's a clause in the docs to address this now.

aantron (commented):
Ok, looks good now! Thanks!

Apart from the two nits, the main remaining question is: do we have license/usage and source information for the OCaml logo used in this example?

By comparison, example `w-one-binary` uses a silly picture of OCaml, and "serves" together with it a [README.md](https://github.com/aantron/dream/tree/master/example/w-one-binary/assets) that says:

> The [camel](camel.jpeg) image is taken from [pixabay.com][source]. The license
states:
> 
> > Free for commercial use
> >
> > No attribution required
>
> [source]: https://pixabay.com/photos/morocco-camel-desert-sand-sahara-5271734/

Since we don't have a separate directory in this example to put a README in, we could just put the corresponding info for the OCaml logo into a different file (or make an `assets` directory like in `w-one-binary`).

jsthomas:
I've added `license_info.md` to explain how the logo used in the example is licensed.

aantron:
Thanks!

PR #141, opened 2021-07-23
Add info about creating local Dream dependency from external dir

outkine:
Add information about how to configure projects that need to reference
the local version of Dream from inside an external directory.


aantron:
Thanks!

PR #146, opened 2021-07-30
Upgrade mirage

dinosaure:
This is an upgrade of the MirageOS support which adds the HTTP (without TLS) support to easily deploy the example locally. However, it seems that `dream-mirage` was not in sync with the core of `dream`. Indeed, `dream-mirage` does not compile anymore and it's weird that the CI did not catch that.

I extended the `README.md` to explain how to deploy the example locally with the `unix` target.

### DIFF
--- a/dream-mirage.opam
+++ b/dream-mirage.opam
@@ -54,6 +54,7 @@ depends: [
   "bigstringaf"
   "digestif" {>= "1.0.0"}
   "lwt"
+  "tls-mirage"
   "lwt_ppx" {>= "1.2.2"}
   "letsencrypt" {>= "0.3.0"}
 ]
--- a/example/m-mirage/README.md
+++ b/example/m-mirage/README.md
@@ -36,7 +36,7 @@ address.
 
 ```sh
 $ opam install mirage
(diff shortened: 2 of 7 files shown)


aantron:
Thanks!

In principle, can the Mirage version run locally with the [development certificate](https://github.com/aantron/dream/tree/master/src/certificate)?

dinosaure:
Hmmhmm it's a bit more complex. You need to have something which can save your certificate **into** the unikernel such as `ocaml-crunch`.

Indeed, the unikernel can not have an access to the host's file-system. Do you want this ability? I can propose something if you want.

aantron:
A development certificate is already compiled into Dream:

https://github.com/aantron/dream/blob/ae3b5d24d35a77c79d718228b1c415ba7db4c6af/src/certificate/dune#L10-L20

It's used internally as `Dream__localhost.certificate` and `Dream_localhost.key`:

https://github.com/aantron/dream/blob/ae3b5d24d35a77c79d718228b1c415ba7db4c6af/src/http/http.ml#L670

PR #148, opened 2021-07-31
Add drop_cookie helper

### LINKED ISSUE #144: Add drop_cookie helper
```ocaml
Dream.drop_cookie name request response =
  Dream.set_cookie name "" request response ~expires:0. ~encrypt:false
```

`Dream.drop_cookie` should have all the same optionals as [`Dream.set_cookie`](https://aantron.github.io/dream/#val-set_cookie), and simply forward them, except that it shouldn't have `~expires`, `~max_age`, or `~encrypt` at all. It can be added right after `set_sookie`:

https://github.com/aantron/dream/blob/ead188384b89322455eb6a6c69541a9ece48e34b/src/pure/inmost.ml#L647

Patil2099:
Fixes #144

### DIFF
--- a/src/pure/inmost.ml
+++ b/src/pure/inmost.ml
@@ -646,9 +646,6 @@ let cookie
 
 let set_cookie
     ?prefix:cookie_prefix
-    ?encrypt:(encrypt_cookie = true)
-    ?expires
-    ?max_age
     ?domain
     ?path
     ?secure


thangngoc89:
I’m not sure if this PR is read my for review or not but this is definitely not adding drop_cookie but rather modifing set_cookie ‘s signature

aantron:
> I’m not sure if this PR is read my for review or not but this is definitely not adding drop_cookie but rather modifing set_cookie ‘s signature

Indeed, thanks @thangngoc89.

@Patil2099 Rather than modifying `set_cookie`, we need to add a new function `drop_cookie` that calls `set_cookie` internally in its implementation.

aantron:
Closing in favor of #169.

PR #149, opened 2021-08-01
Unvendor everything except paf for esy

ulrikstrid:
This should in theory allow us to use Piaf and other things together with Dream if we're using esy (and potentially nix)


aantron:
Thanks!

dinosaure:
Not sure that it's the correct way to do to still be installable _via_ `opam`.

ulrikstrid:
@dinosaure this is on a separate branch, the "real" solution would be if @anmonteiro's work could be upstreamed and/or released.

aantron:
I've since first renamed the vendored libraries, and now replaced them with depending on httpun. Do we still need the `esy-unvendor` branch?

ulrikstrid:
I don't think so. The branch is also very outdated so it likely doesn't make sense to use either way.

PR #152, opened 2021-08-03
Make template example more complex

### LINKED ISSUE #87: Add example showing how to do OCaml control flow in eml templates
...because it can be a bit confusing. It should just show things like

```html
<% items |> List.iter begin fun item -> %>
  <p><%s item %></p>
<% end; %>
```

and/or

```html
%   foos |> List.iter begin fun foo ->
      <p><%s foo %></p>
%   end;
```

which are relatively neat "patterns" for doing control flow, and will help avoid messy errors and taking wrong turns.

cc @tmattio

outkine:
Include if/match statement and for loop. Use different syntax styles to illustrate the range of possibilities. Closes #87.


aantron:
I suggest either adding a new example which is prominently linked from the basic and simplistic template example, or adding snippets to the text of the template example.

This PR makes the basic template example so complex, that I think it becomes frightening to a complete beginner, considering its basic purpose is just to show the absolute most simple thing you can do with the templater, and that it is not scary. We just need some additional tips somewhere to help beginners take the next steps after that, but we don't want to mess up the very first and most basic step, either.

outkine:
Oh, my bad, for some reason I missed the "add new example" part, this makes sense.

outkine:
(this commit addresses #153)

aantron:
Thanks!

Issue #145, opened 2021-07-29
How to run the Mirage example locally?


aantron:
@dinosaure, what would a person new to Mirage (such as myself) do to run the Mirage example locally?

Ideally, the basic Mirage example would show such instructions. It can have a note near the end, giving instructions for gcloud.

dinosaure:
The current example provided by `dream` has some requirement due to the let's encrypt challenge, so it needs for any targets (including `unix`):
- a domain-name (which points to the public IP address)
- at least, the ability to bind `*:80` to do the let's encrypt challenge

I will try to provide something which does not require such things to be able to deploy locally the _unikernel_.

dinosaure:
#146 should fix the issue if it corresponds to what you want. I'm not the best to explain stuffs.

aantron:
As of now, it appears that the Mirage example is still not runnable locally in practice, because it requires a domain and Let's Encrypt.

As mentioned in https://github.com/aantron/dream/pull/146#issuecomment-890515144, Dream has a development certificate already compiled in that can be used (AFAIK) instead of one obtained from Let's Encrypt. It's available under the name `Dream__localhost.certificate`. It's used (internally) by Dream's existing non-Mirage [HTTPS example](https://github.com/aantron/dream/tree/master/example/l-https#files).

I don't know the details, but I initially assume that this certificate can also be used with Mirage.

aantron:
I've since written a new Mirage example, which is much simplified, and I am often running it in order to test Dream's compatibility with Mirage.

The previous Mirage example is now in `w-mirage-letsencrypt` as an archive. I will restore it later once we are working on Dream's Let's Encrypt support.

PR #157, opened 2021-08-13
Be able to launch locally a DreamOS with TLS

dinosaure:
A final fix for #145 with the support of TLS for a local usage. Note that we depends on `tls.0.14.0` now and it seems fine according to my context.


aantron:
Thank you!

PR #158, opened 2021-08-29
Windows: Addresses missing SIGPIPE for Windows

ManasJayanth:
Since SIGPIPE is not found on Windows, this PR adds alternate code path for Windows.

### DIFF
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -813,7 +813,9 @@ let run
     ?(adjust_terminal = true)
     user's_dream_handler =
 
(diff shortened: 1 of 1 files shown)


aantron:
LGTM! Thanks!

I see this is marked as a draft. Is there anything else left to do here?

ManasJayanth:
I wanted to take a closer look again before I requested your attention,  just to be safer. But couldn't find time in the weekend. The patch is small, harmless and seems fine IMO. Marking it ready.

aantron:
Thanks again!

Issue #90, opened 2021-07-02
Dream.flash_message should provide debug mode


thangngoc89:
From my experience adding flash message, it could be extremely frustrating to debug why flash messages aren't rendered at all. 

I wrote this little middleware to help with debugging but this should probably be added to `Dream.flash_message` directly 

```ocaml
type flash = (string * string) list [@@deriving show]

let debug_flash inner_handler request =
  let%lwt response = inner_handler request in
  let local = Dream.get_flash request in
  Dream.log "flash messages: %s" (show_flash local) ;
  Lwt.return response
```

aantron:
I guess there should probably be a `dream.flash` log source. Dream also needs a neat way to set the log level per-source, and then flash messages need documentation about the existence of verbose logging on `dream.flash` at level debug. Would that do?

aantron:
A good place for such documentation might be the not-yet-integrated [flash message example](https://github.com/aantron/dream/tree/809835dcfe655bbe2fd811acae4ddecd6f1800b4/example/b-flash-message), which also needs a README.

aantron:
The flash message middleware should probably also warn if the whole encrypted cookie is more than 4096 bytes long.

PR #159, opened 2021-09-22
Allow log levels to be configured per-source.

jsthomas:
This change proposes an update to `Dream.initialize_log` that allows log levels to be specified for individual log sources.

This is motivated by comments in #90. There, @aantron mentioned the need for a way to configure log level per source, which would be useful for working with logs related to flash messages.

**Example Usage**
```ocaml
let () =
  Dream.initialize_log
    ~level:`Warning
    ~custom_levels:[("dream.flash", `Info)]   (* NEW *)
    ();
  Dream.run
  ...
``` 

The change introduces an optional argument called `custom_levels` that accepts an association list. Sources not named in `~custom_levels` are set to `~level`, so `Dream.initialize_log` retains its original behavior when `~custom_levels` is not used.

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -55,7 +55,7 @@ depends: [
   "conf-libev" {os != "win32"}
   "cstruct" {>= "6.0.0"}
   "dune" {>= "2.7.0"}  # --instrument-with.
-  "fmt" {>= "0.8.7"}  # `Italic.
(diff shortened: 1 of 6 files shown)


jsthomas:
In the interest of simplicity, I'm going to close this PR and open a new one that has the latest changes on the master branch.

Issue #160, opened 2021-09-23
Examples on Apple M1 need an additional resolution in esy. 


pm-mck:
As of today, I found that running the examples in esy on an Apple M1 machine requires this resolution:

```json
  "resolutions": {
    "esy-openssl": "esy-packages/esy-openssl#619ae2d46ca981ec26ab3287487ad98b157a01d1"
  }
```

documented on the esy-openssl project page:

https://github.com/esy-packages/esy-openssl

If you don't do this, the esy build emits a few clang errors which will be difficult for a newcomer to handle.

Should this be added to the examples README or documented / provided in some other manner? Happy to open a pr but don't want to put it in the wrong place.

aantron:
Can you make a PR that adds this to the resolutions of `example/2-middleware` here:

https://github.com/aantron/dream/blob/52d4da207ff2ae51ade431cc88ab26501a69cee7/example/2-middleware/esy.json#L11-L13

From your message, I suppose this will fix the example for you, and choosing this example for the PR will also get the change tested in CI, since that example is part of esy's quick start script that gets run in CI. If everything works, I'll do the busywork of adding this resolution to all the other examples.

@ManasJayanth is it correct to use the resolution this way? I assume that `esy-openssl#619ae2d46ca981ec26ab3287487ad98b157a01d1` will work on all platforms.

sagarjs:
thank you for creating this @pm-mck
I was just about to create this myself though I have no idea how to fix the issue.

I tried adding the line to my esy.json in 2-middleware but I am getting the following error while running 'npx esy'
It looks like permission errors, but if i try manually creating the directory dynlinks as the logs seems to suggest, I do not get an error
what am I getting wrong here?

```
About the OCAML core installation:
        Standard library:      /Users/sagarsodah/.esy/3_____________________________________________________________/i/ocaml-4.12.0-44d44425/lib/ocaml
        Binaries:              /Users/sagarsodah/.esy/3_____________________________________________________________/i/ocaml-4.12.0-44d44425/bin
        Manual pages:          /Users/sagarsodah/.esy/3_____________________________________________________________/i/ocaml-4.12.0-44d44425/man
        Multi-threading type:  posix
    The directory of site-specific packages will be
        site-lib:              /Users/sagarsodah/.esy/3_____________________________________________________________/i/opam__s__ocaml_secondary_compiler-opam__c__4.08.1-1-eca62758/share/ocaml-secondary-compiler/lib
    The configuration file is written to:
[...]

PR #161, opened 2021-09-26
Deploy Fly example

dangdennis:


dangdennis:
Interesting. Looks like the macOS build failed because of the same reason in #160

aantron:
Thank you! Sorry about the delay, I got held up learning about Fly (which seems pretty cool). I'll "digest" this example inside the repo a bit, meaning fully exercise it, add links to it, etc.

PR #163, opened 2021-09-27
set the resolutions for Apple M1 build, remove ocamlfind-secondary for esy build in 2-middleware example

pm-mck:
# Changes

Only for the 2-middleware example:

- [x] Removed `ocamlfind-secondary`. Based on conversations in https://github.com/aantron/dream/issues/160 it is not necessary
- [x] Added appropriate esy-openssl resolution in `esy.json` to enable build on arm64

# Testing

- [x] Tested build and run on x86-64 osx machine
- [x] Tested build and run on arm64 osx machine

### DIFF
--- a/example/2-middleware/esy.json
+++ b/example/2-middleware/esy.json
@@ -5,11 +5,11 @@
     "ocaml": "4.12.x"
   },
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


aantron:
Thank you! I'll go on and apply this to all the examples.

Issue #11, opened 2021-04-08
More optionals for decoupling middlewares


aantron:
The middlewares are internally quite decoupled, but then configured with defaults that make some of them rely on others. Add:

- [ ] Add [`Dream.form`](https://aantron.github.io/dream/#val-form), [`Dream.multipart`](https://aantron.github.io/dream/#val-multipart) [~csrf:false] or [~csrf:my_function] for no or custom CSRF checking, which removes the dependency of the form reader on sessions (because the default CSRF checker uses tokens that are bound to sessions).
- [ ] Add [`Dream.sql_sessions`](https://aantron.github.io/dream/#val-sql_sessions) [~sql:connection] for using a different database connection other than the one supplied by [`Dream.sql_pool`](https://aantron.github.io/dream/#val-sql_pool). This is probably a quite common scenario.

PR #167, opened 2021-10-15
Add csrf option to `Dream.form` and `Dream.multipart`

joseemds:
Part of #11

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -816,7 +816,7 @@ type 'a form_result = [
     activity, or tokens so old that decryption keys have since been rotated on
     the server. *)
 
-val form : request -> (string * string) list form_result promise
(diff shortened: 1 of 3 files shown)


aantron:
Thanks, this looks good, and I'm glad the diff is minimal (no extra whitespace changes, etc.)!

One question: is this motivated by need in your case (i.e. you wanted to use the form helpers without the CSRF check)?

joseemds:
Hi! Yes it is, i am playing with Dream and have an endpoint that expects a `Content-Type:  application/x-www-form-urlencoded`, but couldn't consume it with `curl` neither test it with `Dream.test`, so i searched a way of disable the `csrf` and found the issue

aantron:
Great, thanks for telling. Helps actually motivate it (as I was only suspecting it would be useful, hadn't actually observed a use!)

PR #169, opened 2021-10-21
feat: Add drop_cookie helper

### LINKED ISSUE #144: Add drop_cookie helper

shawn-mcginty:
Fixes #144

I tried also building the docs but that seemed to fail for me, but I did add `Dream.drop_cookie` to the cookie example.

### DIFF
--- a/src/pure/inmost.ml
+++ b/src/pure/inmost.ml
@@ -702,3 +702,35 @@ let set_cookie
   in
 
   add_header "Set-Cookie" set_cookie response
+
+let drop_cookie
+    ?prefix:cookie_prefix
+    ?domain
+    ?path
+    ?secure
+    ?(http_only = true)
+    ?same_site
+    name
+    request
+    response =
+  let set_cookie = set_cookie ~http_only:http_only ~expires:0. ~encrypt:false in
+
+  let set_cookie = match cookie_prefix with
+    | Some p -> set_cookie ~prefix:p
+    | None -> set_cookie ~prefix:None in
+  
+  let set_cookie = match path with
+    | Some p -> set_cookie ~path:p
+    | None -> set_cookie ~path:(Some (prefix request)) in
+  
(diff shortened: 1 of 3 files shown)


src/pure/inmost.ml
```diff
@@ -702,3 +702,35 @@ let set_cookie
...
+    response =
+  let set_cookie = set_cookie ~http_only:http_only ~expires:0. ~encrypt:false in
+
+  let set_cookie = match cookie_prefix with
+    | Some p -> set_cookie ~prefix:p
+    | None -> set_cookie ~prefix:None in
+  
+  let set_cookie = match path with
+    | Some p -> set_cookie ~path:p
+    | None -> set_cookie ~path:(Some (prefix request)) in
```
aantron, line 725: Most of the arguments can be simply forwarded to `set_cookie`, e.g.

```
set_cookie (* ... *) ?path(* ... *)
```

The `?secure` syntax forwards an optional argument to an optional argument, so you don't have to pattern match on them and then build up a `set_cookie` closure gradually, as here.

I'll take care of tweaking this after merge.

aantron:
Thanks!

src/pure/inmost.ml
shawn-mcginty: Wow thanks, I was looking for a feature like this but I could not find it anywhere.

PR #171, opened 2021-11-03
Allow log levels to be configured per-source.

jsthomas:
This change updates the logging API to allow log levels for sub-logs to be customized. This is a revision of #159, rebased on the latest in master for easier review. The change is motivated by issue #90; configurable log levels would assist with debug and warning logs for flash messages.

The change adds an optional argument to set the level of a `sub-log` at creation:
```ocaml
Dream.sub_log : ?level:[< log_level ] -> string -> sub_log
```
as well as a new function to set the log level afterward:
```ocaml
Dream.set_log_level : string -> [< log_level ] -> unit
```

**Example Usage**
```ocaml
let () =
  Dream.initialize_log ~level:`Warning ();
  Dream.set_log_level "dream.flash" `Info;
  let app_log = Dream.sub_log ~level:`Info "my_app" in 
  Dream.run
  ...
  (* Now log messages on my_app and dream.flash will show if they are of info-level or higher.
     Other messages must be warning-level or higher to be displayed.*)
```

PR #173, opened 2021-11-10
Issue 90: Add additional instrumentation for flash messages.

jsthomas:
This change addresses #90. It introduces a new debug log point that records the current state of the flash message cookie on every request. Using the custom log levels introduced in #171, developers can easily set the `dream.flash` log source to debug level to trace their flash messages. The flash message example (`w-flash`) has been updated to illustrate how to do this.

On some browsers (for example, Firefox) cookies are limited to a maximum of 4096 bytes; above this limit, a browser may drop the cookie. Since this can prevent flash messages from appearing, I've also included a warning log point to alert developers when their message content is too big.

### DIFF
--- a/example/w-flash/README.md
+++ b/example/w-flash/README.md
@@ -48,6 +48,8 @@ into cookies and reads them back out:
 
 ```ocaml
 let () =
+  Dream.initialize_log ~level:`Info ();
+  Dream.set_log_level "dream.flash" `Debug;
   Dream.run
   @@ Dream.logger
   @@ Dream.memory_sessions
@@ -73,6 +75,16 @@ let () =
 
   ]
   @@ Dream.not_found
+```
+
+The example configures a custom log level for flash messages using
+`Dream.set_log_level`. Setting this to "debug" means the server logs
+will display a log point summarizing the flash messages on every
+request, like this:
+
+```
+10.11.21 01:48:21.629       dream.log  INFO REQ 3 GET /result ::1:39808 Mozilla/5.0 (X11; Linux x86_64; rv:78.0) Gecko/20100101 Firefox/78.0
+10.11.21 01:48:21.629     dream.flash DEBUG REQ 3 Flash messages: Info: Some Message
 ```
 
 <pre><code><b>$ cd example/w-flash</b>
--- a/example/w-flash/README.md
+++ b/example/w-flash/README.md
@@ -48,7 +48,6 @@ into cookies and reads them back out:
 
 ```ocaml
 let () =
-  Dream.initialize_log ~level:`Info ();
   Dream.set_log_level "dream.flash" `Debug;
   Dream.run
   @@ Dream.logger
--- a/example/w-flash/flash.eml.ml
+++ b/example/w-flash/flash.eml.ml
@@ -33,8 +33,8 @@ let () =
     Dream.post "/"
       (fun request ->
         match%lwt Dream.form request with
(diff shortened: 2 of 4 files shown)


example/w-flash/README.md
```diff
@@ -48,6 +48,8 @@ into cookies and reads them back out:
 
 ```ocaml
 let () =
+  Dream.initialize_log ~level:`Info ();
```
aantron, line 51: AFAIK this call shouldn't be necessary. Is it? If `set_log_level` requires it, `set_log_level` should be altered to do a lazy internal `initialize_log`.

example/w-flash/flash.eml.ml
```diff
@@ -31,8 +33,8 @@ let () =
     Dream.post "/"
       (fun request ->
         match%lwt Dream.form request with
-        | `Ok ["text", text] ->
-          let () = Dream.put_flash "Info" text request in
+          | `Ok ["text", text] ->
+            let () = Dream.put_flash "Info" text request in
```
aantron, line 37: Why have these lines had an indentation-only change?

Also, where does this kind of indentation come from? It is wasteful of the left margin.

src/middleware/flash.ml
```diff
@@ -76,3 +56,40 @@ let put_flash category message request =
...
+
+let flash_messages inner_handler request =
+  let outbox = ref [] in
+  let request = Dream.with_local storage outbox request in
+  let%lwt response = inner_handler request in
+  let entries = List.rev !outbox in
+  let existing = Dream.cookie flash_cookie request in
+  log.debug (fun log ->
+      let current =
+        flash request
```
aantron, line 68: Shouldn't this display the contents of `outbox`, rather than reading the cookie, which AFAIK is the flash messages returned in the request? If you'd like to display incoming flash messages as I think this is doing, I think it should be done before the call to `inner_handler`.

jsthomas: In the issue, @thangngoc89 wrote that it was "frustrating to debug why flash messages aren't rendered". I thought it made sense to log the flash messages from the current request's cookie because those are the ones available for rendering a response. This also appears to match the behavior of the example code in the issue description. 

I've fixed the ordering with `inner_handler`; that was an oversight on my part.

example/w-flash/README.md
jsthomas: When I tested this against the `w-flash` example it did appear to be necessary; unfortunately I don't fully understand why. I've updated `set_log_level`.

src/middleware/flash.ml
aantron: The question is, in the original ordering, why read the cookie on the *request* (`flash request`), rather than reading `outbox`? Aren't the messages in the outbox the ones that will be put in the response?

aantron: Ah okay, I understood you. I got double-negative-confused by myself :) Okay, I now think this is doing what #90 was asking, and the ordering makes sense now.

aantron: Could you move the debug print all the way to the top of the middleware? It doesn't seem to depend on any of the values bound before it now, and having them bound there makes it seem like it might depend. It's a clean print of the flash messages being returned to the server. It will especially make it clear that the logging does not depend on the rebinding of `request` to one that has the `outbox` attached.

aantron:
Okay, I realized I have only trivial nits left, so I merged. Thanks for doing this with the several steps it took!

PR #175, opened 2021-11-15
Fix `make depends` command on DreamOS

dinosaure:
I wrote the `config.ml` too fast. We must use `sublibs` for sub-libraries to be able to generate a well-formed `opam` file.


aantron:
Thanks!

PR #182, opened 2021-11-29
Update gluten & tls

dinosaure:
A gently PR to update `gluten` and use the last version of `tls.0.15.0`.


aantron:
Thanks! This looks like a variant of https://github.com/anmonteiro/gluten/pull/24. @anmonteiro, what do you think about merging that PR? We could then point Dream to the merged commit. Alternatively, we could point Dream to @hannesm's fork that the PR originates from, or 0c9341a64ee7432c7a3e1a5e97b4012fee2775c2 as proposed by @dinosaure here in this PR, but I'm pretty wary of generating even more stack forks and mystery vendored branches than Dream already is using :)

aantron:
To unblock this, I pulled the Gluten commit into aantron/gluten, branch `fork`. I'll repoint the submodule to whatever gets merged into upstream Gluten later.

PR #186, opened 2021-12-20
Move dependencies of dream-pure to that module

fkrull:
This was necessary for me to use Dream from git with esy because was complaining about missing dependencies in dream-pure otherwise (quite reasonably from what I can tell).

### DIFF
--- a/dream-pure.opam
+++ b/dream-pure.opam
@@ -18,9 +18,11 @@ depends: [
   "dune" {>= "2.7.0"}  # --instrument-with.
   "hmap"
   "lwt"
+  "lwt_ppx" {>= "1.2.2"}
   "ocaml" {>= "4.08.0"}
   "ptime" {>= "0.8.1"}  # Ptime.weekday.
   "uri" {>= "4.2.0"}
+  "multipart_form" {>= "0.3.0"}
 
   # Testing, development.
   "alcotest" {with-test}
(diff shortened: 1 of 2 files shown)


aantron:
Thanks for the PR! I indeed had a TODO about double-checking the dependencies of the new packages, and you have addressed it! :)

`dream-pure` in fact should not depend on `multipart_form`, that was a [stray dependency left over in the `dune` file](https://github.com/aantron/dream/commit/4fbb0c56f4e64ff83d41324e7720ef4d3effc039) from an earlier iteration of `dream-pure`. I removed it in `master`.

As for `lwt_ppx`, `dream-pure` in fact does not need it at this exact moment. However, it does seem best to have it as a dependency, as this PR adds, since refactoring might introduce `let%lwt` somewhere in `dream-pure`, and it's not worth it to constantly remove and add `lwt_ppx` preprocessing based on such refactoring noise, especially given the rest of Dream does use `lwt_ppx`.

I also re-added `lwt_ppx` as a dependency back to `dream.opam`, because package `dream` also directly depends on `lwt_ppx`. I'll also add it to the new `dream-httpaf` package.

aantron:
> As for `lwt_ppx`, `dream-pure` in fact does not need it at this exact moment.

(Meaning the code itself does not need it right now. It is still mentioned in the `dune` file and causes the errros you observed if not listed as a dependency in `dream-pure.opam`, as done by this PR.)

PR #200, opened 2022-02-07
Prefix vendored libraries

### LINKED ISSUE #198: Prefix the vendored libraries
In a recent project, I met various issues due to the forked libraries vendored by `dream`.
Indeed, it was really hard to have the latest versions of many other dependencies due to conflicts between their own dependencies and the dependencies of `dream`.

Indeed, when a library is vendored by `dream` but is also required by another dependency, `opam` is unable to understand which one of them he should use.

I have temporarily solved this issue by forking `dream` and all it's vendored libraries, and prefixing everyone of them following the format `dream-*`. This is not optimal, but it is the only option that I could think about.

Could it be also done on the main repository, so the problem would be permanently solved ?

aantron:
@clecat, thanks for the patch and for uploading your branches of http/af, etc. I imported your prefixed branches into my forks and pointed Dream to them, then applied your patch, doing a lot of work on it to merge it with current Dream `master`, and also getting some missed parts (`dream-websocket`).

There is still a bit of work left. I had created an independent [`dream-httpaf` library](https://github.com/aantron/dream/blob/257f0e0d41828d7eefa1e41f2de05dbb8ef9ed7b/dream-httpaf.opam), which contains the vendored http/af stack. Its name now conflicts with the prefixed `dream-httpaf`. I'm thinking to rename the prefixed `httpaf` to `dream_httpaf_`. This probably requires a change back in the vendored fork of httpaf, so that the main module is `Dream_httpaf_`. @clecat, could you resume your workflow (since I would have to work it out), and rename `Dream_httpaf` to `Dream_httpaf_` until `make` succeeds? I'll then get back to this PR and finish off the merge. You can pull the code in this PR from branch `prefix-vendored`.

Fixes #198.

cc @dinosaure

### DIFF
--- a/src/http/adapt.ml
+++ b/src/http/adapt.ml
@@ -68,22 +68,22 @@ let forward_body_general
 
 let forward_body
     (response : Message.response)
-    (body : Httpaf.Body.Writer.t) =
+    (body : Dream_httpaf.Body.Writer.t) =
 
   forward_body_general
     response
-    (Httpaf.Body.Writer.write_string body)
-    (Httpaf.Body.Writer.write_bigstring body)
-    (Httpaf.Body.Writer.flush body)
-    (fun _code -> Httpaf.Body.Writer.close body)
+    (Dream_httpaf.Body.Writer.write_string body)
+    (Dream_httpaf.Body.Writer.write_bigstring body)
+    (Dream_httpaf.Body.Writer.flush body)
+    (fun _code -> Dream_httpaf.Body.Writer.close body)
 
 let forward_body_h2
     (response : Message.response)
-    (body : [ `write ] H2.Body.t) =
+    (body : [ `write ] Dream_h2.Body.t) =
 
   forward_body_general
     response
-    (H2.Body.write_string body)
-    (H2.Body.write_bigstring body)
-    (H2.Body.flush body)
-    (fun _code -> H2.Body.close_writer body)
+    (Dream_h2.Body.write_string body)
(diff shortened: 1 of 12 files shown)


clecat:
Thank you, I will work on it as soon as possible

hannesm:
Dear @aantron and @clecat, I tried to used dream in mirage together with a git implementation, and stumble upon this issue. I tried for some hours to get the module name collisions sorted without any success (now some code in mirage attempts to use `Dream__pure.Inmost` which does not seem to exist anymore).

Is there a path forward? E.g. could the vendored libs (in src/vendor) just not expose any public_name? I'd be keen if this works smoothly in the end.

hannesm:
@ulrikstrid interesting -- could you point me to the branch / commits? :)

ulrikstrid:
Here is my branch: https://github.com/ulrikstrid/dream/tree/piaf-compat

Here is the patch in the nix overlays: https://github.com/anmonteiro/nix-overlays/blob/master/ocaml/dream/upload.patch
Used [here](https://github.com/anmonteiro/nix-overlays/blob/master/ocaml/dream/httpaf.nix) and [here](https://github.com/anmonteiro/nix-overlays/blob/master/ocaml/dream/default.nix)

And here is where it's used in a esy project: https://github.com/marigold-dev/deku/blob/main/esy.json#L86-L111

Hopefully it helps @hannesm, this uses @anmonteiro's packages of httpaf etc

PR #203, opened 2022-02-14
Split content-type header value

yawaramin:
Some clients add a suffix to the value, e.g.
`"application/x-www-form-urlencoded; charset=UTF-8"`. In this case we want to
throw away the extra data before checking the content-type.

### DIFF
--- a/src/server/form.ml
+++ b/src/server/form.ml
@@ -5,6 +5,7 @@
 
 
 
+module Formats = Dream_pure.Formats
 module Message = Dream_pure.Message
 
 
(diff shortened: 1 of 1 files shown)


yawaramin:
Another option would be:

```ocaml
match String.sub ct 0 33 with
| "application/x-www-form-urlencoded" -> ...
```

Let me know if you prefer that.

aantron:
Thank you!

> Another option would be:
> 
> ```ocaml
> match String.sub ct 0 33 with
> | "application/x-www-form-urlencoded" -> ...
> ```
> 
> Let me know if you prefer that.

It looks like either version would allocate. This version allocates less, but what we really would want from an allocations point of view is a `String.startswith` in the standard library.

PR #205, opened 2022-02-22
Update router exemple for alpha3

oteku:
Update of documentation and exemple regarding `Dream.router` in alpha3


ulrikstrid:
CI failure seems to be unrelated

aantron:
Many thanks! Looks like I missed a few examples when updating them. Especially thankful for finding the ones in the `.mli` file.

PR #210, opened 2022-03-07
Add camlp-streams dep for 5.0.0

### LINKED ISSUE #208: `Stream` module no longer available in OCaml 5
Hi :)) Thanks for the great library! 

I'm doing some work with OCaml 5 and since https://github.com/ocaml/ocaml/pull/10896 the `Stream` library now needs to be explicitly pulled in via https://github.com/ocaml/camlp-streams (which I did quickly https://github.com/patricoferris/dream/commit/21e981fdc500f8728e984faaf1fef9bff0d26cfe) but I wasn't sure if this is the approach you wanted to go for (maybe vendoring your own `Stream` for `eml` or something like that?).

Anyway just thought I'd report it as I couldn't see any other issues/PRs open for it.

patricoferris:
This PR adds a dependency on [camlp-streams](https://github.com/ocaml/camlp-streams) after `Stream` was removed from standard library of the compiler for OCaml 5. This fixes #208.

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -49,6 +49,7 @@ maintainer: "Anton Bachin <antonbachin@yahoo.com>"
 depends: [
   "base-unix"
   "bigarray-compat"
+  "camlp-streams"
   "caqti" {>= "1.6.0"}  # https://github.com/aantron/dream/issues/44.
   "caqti-lwt"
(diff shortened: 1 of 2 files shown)


aantron:
Thank you!

Issue #204, opened 2022-02-20
Add `?trim:bool` param to `Dream.query`?


yawaramin:
Or perhaps even just trim the values automatically?

E.g. some clients might send queries like `?q=` (with no value) or `?q= ` (single space character as the value).

In these cases perhaps we should evaluate to `None` for `Dream.query ~trim:true req "q"`?

aantron:
Would you be willing to make a PR for this, calling the option `?filter_empty`?

tatchi:
@yawaramin did you start working on that one? I haven't checked the code yet but that looks like something that _shouldn't_ be too difficult to implement so I would be happy to try :)

yawaramin:
@tatchi I had not, got distracted by other things 😅 Please feel free!

aantron:
@yawaramin what was the original context for this? What client was sending `?q=` or `?q= `, and in what circumstance? Is there some kind of inconsistency between clients in this behavior?

aantron:
I'm going to close this as something very specific to a certain use case (which may not exist) and easily implementable by Dream's users when needed.

yawaramin:
Now with typed form decoding/validation in [dream-html](https://yawaramin.github.io/dream-html/dream-html/Dream_html/Form/index.html), we will be able to use the same functionality for decoding queries into typed values and enforcing constraints like 'must be nonempty'. Obviously for the simple case above it's overkill, but just for demonstration purposes:

```ocaml
open Dream_html.Form

let nonempty = ensure "error.expected.nonempty" (( <> ) "") required string

let handler request =
  let q = validate (nonempty "q") (Dream.all_queries request) in
  ...
```

aantron:
Looks neat!

PR #212, opened 2022-03-10
add a filter_empty optional param to Dream.query

tatchi:
My attempt at implementing #204

@aantron on a side note, I'm using vscode together with `ocaml-lsp-server` and the official OCaml extension. I wanted to format I've edited but it leads to lots of changes 😅 Do you manually format the code?

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -414,7 +414,7 @@ https://github.com/aantron/dream/issues
 (** Replaces the path. See {!Dream.val-path}. *)
 (**/**)
 
-val query : request -> ?filter_empty:bool -> string -> string option
+val query : ?filter_empty:bool -> request -> string -> string option
 (** First query parameter with the given name. See
     {{:https://tools.ietf.org/html/rfc3986#section-3.4} RFC 3986 §3.4} and
     example
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -414,7 +414,7 @@ https://github.com/aantron/dream/issues
 (** Replaces the path. See {!Dream.val-path}. *)
 (**/**)
 
-val query : request -> string -> ?filter_empty:bool -> unit -> string option
+val query : request -> ?filter_empty:bool -> string -> string option
 (** First query parameter with the given name. See
     {{:https://tools.ietf.org/html/rfc3986#section-3.4} RFC 3986 §3.4} and
     example
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -414,7 +414,7 @@ https://github.com/aantron/dream/issues
 (** Replaces the path. See {!Dream.val-path}. *)
 (**/**)
 
-val query : request -> string -> string option
+val query : request -> string -> ?filter_empty:bool -> unit -> string option
 (** First query parameter with the given name. See
     {{:https://tools.ietf.org/html/rfc3986#section-3.4} RFC 3986 §3.4} and
     example
--- a/src/server/query.ml
+++ b/src/server/query.ml
@@ -28,7 +28,7 @@ let all_queries request =
   |> snd
   |> Formats.from_form_urlencoded
 
-let query request name ?(filter_empty = false) () =
+let query request ?(filter_empty = false) name =
   match (List.assoc_opt name (all_queries request), filter_empty) with
   | None, _ -> None
   | Some v, false -> Some v
--- a/src/server/query.ml
+++ b/src/server/query.ml
@@ -28,8 +28,11 @@ let all_queries request =
   |> snd
   |> Formats.from_form_urlencoded
 
-let query request name =
-  List.assoc_opt name (all_queries request)
+let query request name ?(filter_empty = false) () =
+  match (List.assoc_opt name (all_queries request), filter_empty) with
+  | None, _ -> None
+  | Some v, false -> Some v
+  | Some v, true -> if String.trim v = "" then None else Some v
 
 let queries request name =
   all_queries request
--- a/src/server/query.ml
+++ b/src/server/query.ml
@@ -28,7 +28,7 @@ let all_queries request =
   |> snd
   |> Formats.from_form_urlencoded
 
-let query request ?(filter_empty = false) name =
+let query ?(filter_empty = false) request name =
   match (List.assoc_opt name (all_queries request), filter_empty) with
   | None, _ -> None
   | Some v, false -> Some v
--- a/src/server/query.ml
+++ b/src/server/query.ml
@@ -29,10 +29,13 @@ let all_queries request =
   |> Formats.from_form_urlencoded
 
 let query ?(filter_empty = false) request name =
-  match (List.assoc_opt name (all_queries request), filter_empty) with
-  | None, _ -> None
-  | Some v, false -> Some v
-  | Some v, true -> if String.trim v = "" then None else Some v
+  match List.assoc_opt name (all_queries request) with
+  | None -> None
+  | Some v as result ->
+    if filter_empty && String.trim v = "" then
+      None
+    else
+      result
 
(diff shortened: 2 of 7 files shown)


src/server/query.ml
```diff
@@ -28,8 +28,11 @@ let all_queries request =
   |> snd
   |> Formats.from_form_urlencoded
 
-let query request name =
-  List.assoc_opt name (all_queries request)
+let query request name ?(filter_query = false) () =
+  match (List.assoc_opt name (all_queries request), filter_query) with
+  | None, _ -> None
+  | Some v, false -> Some v
+  | Some v, true -> if String.trim v = "" then None else Some v
```
tatchi, line 35: It will `trim` multiple occurrences of spaces. That's not what was described in the original issue but if we only check for one empty space, `" "` will give `None` while multiple spaces will result in `Some("  ")`. It looks a bit weird to me but I'm definitely open to changing it if you prefer.

test/expect/pure/formats/query/query.ml
```diff
@@ -5,22 +5,30 @@
...
+  query ~filter_query:false "c" "a=b&c=d";
+  query ~filter_query:false "a b" "a+b=c";
+  query ~filter_query:false "a" "a";
+  query ~filter_query:true "a" "a";
+  query ~filter_query:true "a" "a ";
+  query ~filter_query:true "a" "a  ";
+  query ~filter_query:true "a" "a=";
+  query ~filter_query:true "a" "a= ";
+  query ~filter_query:true "a" "a=  ";
+  query ~filter_query:true "a" "a= &a=b";
```
tatchi, line 31: That one currently gives `None`. Is that what we want?

src/dream.mli
```diff
@@ -414,7 +414,7 @@ https://github.com/aantron/dream/issues
 (** Replaces the path. See {!Dream.val-path}. *)
 (**/**)
 
-val query : request -> string -> string option
+val query : request -> string -> ?filter_query:bool -> unit -> string option
```
ulrikstrid, line 417: If you move the optional parameter to the left you can skip the final unit argument. Then you don't have to change the examples. 🙂

tatchi: Beautiful, thanks! 🤩

src/dream.mli
```diff
@@ -414,7 +414,7 @@ https://github.com/aantron/dream/issues
 (** Replaces the path. See {!Dream.val-path}. *)
 (**/**)
 
-val query : request -> string -> string option
+val query : request -> ?filter_empty:bool -> string -> string option
```
aantron, line 417: Could you move the optional fully to the left?

src/server/query.ml
```diff
@@ -28,8 +28,11 @@ let all_queries request =
   |> snd
   |> Formats.from_form_urlencoded
 
-let query request name =
-  List.assoc_opt name (all_queries request)
+let query request ?(filter_empty = false) name =
+  match (List.assoc_opt name (all_queries request), filter_empty) with
```
aantron, line 32: I recommend rewriting this as

```ocaml
match List.assoc_opt name (all_queries request) with
| None -> None
| Some v as result ->
  if filter_empty then
    result
  else
    if String.trim v = "" then
      None
    else
      result
```

This avoids unconditionally allocating `Some` options every time there is a `Some v` match. Also, it avoids mixing in the `filter_empty` check into all branches, when it is only relevant for the `Some v` branch.

Later, if actually optimizing, we may want to replace `String.trim` by a scan of the string that does not allocate an intermediate string potentially almost as long as `v`, only for us to compare it with `""`.

test/expect/pure/formats/query/query.ml
```diff
@@ -5,8 +5,8 @@
 
 
 
-let query name string =
-  Dream.query (Dream.request ~target:("/?" ^ string) "") name
+let query name ?(filter_empty) string =
```
aantron, line 8: The parentheses are not necessary here.

test/expect/pure/formats/query/query.ml
```diff
@@ -21,6 +21,14 @@ let%expect_test _ =
...
   query "c" "a=b&c=d";
   query "a b" "a+b=c";
+  query "a" "a";
+  query ~filter_empty:true "a" "a";
+  query ~filter_empty:true "a" "a ";
+  query ~filter_empty:true "a" "a  ";
+  query ~filter_empty:true "a" "a=";
+  query ~filter_empty:true "a" "a= ";
+  query ~filter_empty:true "a" "a=  ";
+  query ~filter_empty:true "a" "a= &a=b";
```
aantron, line 31: Can you add `query ~filter_empty:true "a" "a= b "` and any other tests that use `~filter_empty` but should *not* produce `None`?

aantron (commented):
Thank you! Just have a few notes and nits.

aantron:
(But note, if you'd like to format some code you're adding, Dream does have an [`.ocamlformat`](https://github.com/aantron/dream/blob/master/.ocamlformat) in its repo that produces some approximation of a modern, diff-friendly, and Web-oriented style, to the extent that's currently possible with Ocamlformat.)

test/expect/pure/formats/query/query.ml
aantron: Yes. [`Dream.query`](https://aantron.github.io/dream/#val-query):

> First query parameter with the given name.

[`Dream.queries`](https://aantron.github.io/dream/#val-queries) can be used to get the result `[" ", "b"]` here.

src/server/query.ml
tatchi: > This avoids unconditionally allocating Some options every time there is a Some v match. Also, it avoids mixing in the filter_empty check into all branches, when it is only relevant for the Some v branch.

Oh, that makes sense, thanks for the explanation 😊 I think the conditions should be inverted in your example. It should be:

```ocaml
match List.assoc_opt name (all_queries request) with
| None -> None
| Some v as result ->
  if filter_empty then
    if String.trim v = "" then
      None
    else
      result
  else
    result
```

Which I've simplified to:

```ocaml
match List.assoc_opt name (all_queries request) with
| None -> None
| Some v as result ->
  if filter_empty && String.trim v = "" then
    None
  else
    result
```

I hope that makes sense.

tatchi: > Later, if actually optimizing, we may want to replace String.trim by a scan of the string that does not allocate an intermediate string potentially almost as long as v, only for us to compare it with "".

Interesting, would something like the following be reasonable ?

```ocaml
(* taken from https://github.com/ocaml/ocaml/blob/trunk/stdlib/string.ml#L97 *)
let is_space = function
  | ' ' | '\012' | '\n' | '\r' | '\t' -> true
  | _ -> false
  
let is_empty s =
  let len = String.length s in
  if len = 0 then
    true
  else
    let rec loop len i =
      if i < len && is_space s.[i] then
        loop len (i + 1)
      else
        false
    in
    loop len 0
```

Or without the inner `loop` function:

```ocaml
let is_empty s =
  let len = String.length s in
  if len = 0 then
    true
  else
    let i = ref 0 in
    let contains_only_spaces = ref true in
    while !contains_only_spaces && !i < len do
      if is_space s.[!i] then
        i := !i + 1
      else
        contains_only_spaces := false
    done;
    !contains_only_spaces
```

tatchi:
Nice to see you back working on this project @aantron 🤗😍

aantron:
Thank you :)

aantron:
After the huge delay (apologies!) I think this behavior is too niche to support directly in Dream, given a user can easily write their own wrapper to deal with the inconveniences of their client's behavior. I'm going to close this, but we can resurrect it if we have claims otherwise. Sorry and thank you!

PR #219, opened 2022-04-29
Add support for listening to unix domain sockets

### LINKED ISSUE #218: Support for unix domain sockets?
In Go, it's quite easy to start a server listening on unix domain sockets:

https://pkg.go.dev/net#Listen

Would it be possible to add this to Dream as well?

samoht:
Fix #218

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -2193,6 +2193,7 @@ val catch : (error -> response promise) -> middleware
 val run :
   ?interface:string ->
   ?port:int ->
+  ?socket_path:string ->
   ?stop:unit promise ->
   ?error_handler:error_handler ->
   ?tls:bool ->
@@ -2252,6 +2253,7 @@ val run :
 val serve :
   ?interface:string ->
   ?port:int ->
+  ?socket_path:string ->
   ?stop:unit promise ->
   ?error_handler:error_handler ->
   ?tls:bool ->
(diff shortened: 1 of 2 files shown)


src/dream.mli
```diff
@@ -2193,6 +2193,7 @@ val catch : (error -> response promise) -> middleware
 val run :
   ?interface:string ->
   ?port:int ->
+  ?socket_path:string ->
```
askvortsov1, line 2196: I wonder if it would be a cleaner API to offer separate functions for serving on a port (which would likely be used more) vs on a "network address", which could correspond to the `network` type defined elsewhere in this PR? Feels like we shouldn't accept multiple config parameters if some of them could get cancelled out. Possibly `run_sock` and `serve_sock`?

Doesn't look like `interface` is needed if listening on a unix socket either; pretty sure `is_localhost` would just need to be adapted to return `true` if on a socket.

src/http/http.ml
```diff
@@ -379,13 +379,15 @@ let built_in_middleware error_handler =
     Catch.catch (Error_handler.app error_handler);
   ]
 
-
+type network =
```
askvortsov1, line 382: Why not reuse `Unix.sockaddr`?

src/dream.mli
ewanas: Maybe a `socket_source` instead of `interface`, `port` or `socket_path` can be passed. `Inet` or `Unix` being possible kinds of sources. Although that will make the run function slightly more cumbersome, but the implementation can just default to `Inet(defaultInterface, defaultPort)`.

aantron:
Thank you! I'll add docs in a follow-on commit.

PR #223, opened 2022-05-26
Fix build failure due to label warning

laaksonel:
Compilation fails due to unlabeled argument `p`

```
rescript: [6/6] client/client.cmj
FAILED: client/client.cmj

  Warning number 6 (configured as error)
  .../dream/example/w-fullstack-rescript/client/client.res:14:11-29

  12 │     let p = document->Document.createElement("p")
  13 │     p->Element.setInnerText(text)
  14 │     body->Element.appendChild(p)
  15 │   }
  16 │ }

  label child was omitted in the application of this function.
```

### DIFF
--- a/example/w-fullstack-rescript/README.md
+++ b/example/w-fullstack-rescript/README.md
@@ -58,7 +58,7 @@ let () = {
 
     let p = document |> Document.createElement("p")
     p->Element.setInnerText(text)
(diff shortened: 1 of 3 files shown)


aantron:
Thank you! This is due to a change in `rescript-webapi` 0.5.0, so I've also added that as a constraint to `package.json`.

Issue #227, opened 2022-06-10
Consider a more Unix-y interface for dream_eml


tmattio:
The current `dream_eml` CLI takes a `FILE` argument, infers the output filename from it, and writes this file in the same directory as the input file.

This has been problematic in the past when I wanted to put my eml files in a separate directory.
The interface also makes it impossible to create a new dune `dialect` as the preprocessor is expected to write to stdout.

I think a more Unix-y interface for the `dream_eml` CLI would make its usage a bit more flexible.

What do you think of allowing it to read from stdin and write to stdout? Should these be the default?

aantron:
I agree with this in principle and see the advantages. However, `eml` needs a `FILE` argument to generate locations in its output that refer to the original source file, i.e. `#22 src/foo/bar.ml`. That's why the command line is not Unix-y to begin with. Do you really need reading from STDIN, or just writing to STDOUT, like in @tcoopman's PR? An interface where `eml` always reads from STDIN would be a bit awkward, since it would look like `dream_eml FILE.eml < FILE.eml > FILE.ml`.

PR #228, opened 2022-06-10
add stdout as output from dream_eml

tcoopman:
quick and dirty implementation for #227 

this should work with this `dialect` in `dune-project`

```
(dialect
 (name eml)
 (implementation
  (extension eml)
  (preprocess (run ./main.exe --stdout %{input-file})))
 (interface
  (extension emli)
  (format (run cat)))
 )
```

I only tested this very briefly but it seems to work fine. I think the `interface` stanza is required, but that will never been needed I guess?

If this is something that is worth doing, let me know and then I can maybe look into rewriting some examples as well to use the dialect?

### DIFF
--- a/src/eml/eml.ml
+++ b/src/eml/eml.ml
@@ -782,7 +782,7 @@ end
 
 
 
-let process_file (input_file, location, syntax) =
+let process_file (input_file, location, syntax, std_out) =
   let reason, extension = match syntax with
   | `OCaml -> (false, ".ml")
   | `Reason -> (true, ".re")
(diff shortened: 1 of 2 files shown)


gpetiot:
This would be very useful for ocaml.org! (https://github.com/ocaml/ocaml.org/pull/1181)

aantron:
@tmattio, @gpetiot, @emillon, I've opened two issues on Dune (https://github.com/ocaml/dune/issues/7776, https://github.com/ocaml/dune/issues/7777) about improving the ergonomics, and I think I will merge this regardless. I might tweak the command line before release, and deprecate the "old" instructions for integrating eml into projects. Very glad to learn about `(dialect)`.

aantron:
Thank you! I will update the docs to describe this in a follow-on commit.

aantron:
Since Dream's examples recommend `.eml.ml` files, documenting this is blocked on https://github.com/ocaml/dune/issues/7777 ("Support multiple extensions in (dialect) patterns") in Dune, which will probably come out in the next release. This is already usable from Dream `master` with single extensions, like `.eml`, as found in the Ocaml.org sources.

PR #229, opened 2022-06-10
Fix tests on Linux

beajeanm:
* The pin of reason is no longer necessary as reason 3.8 supports ocaml 4.13
* Fix a higher bond to caqti. Caqti 1.8 introduces new deprecations that would
  require code change.
* The behavior of `Printexec.to_string (Failure "msg")` has changed from 4.09
  to 4.10. It used to be `(Failure "write to a read-only stream")` but it's now
  `Failure ("write to a read-only stream")`. Since we already know the exception's
  type, I figure it would be enough to match on the message itself.


aantron:
I cherry-picked the parts of this that are still up-to-date as 14093e16ccede8fd29f0b4b6cdd0933e1e32cc93.

aantron:
Thank you!

PR #241, opened 2022-10-31
Point to patched ocaml-h2.

novemberkilo:
Hi - this follows from working on getting `dream` to build on OCaml 5.

Please see [this build failure](http://check.ocamllabs.io/log/1666790526-445ff10841f25868d96c33b756783bfa4ceb4f44/5.0+alpha-repo/partial/dream.1.0.0~alpha4) -- essentially we need a patch on `httpaf` which in turn needs a patch on `h2` to fix the build.
 
Currently pointing h2 to my fork of it that builds on OCaml 5. 

Will update once https://github.com/aantron/ocaml-h2/pull/1 is merged.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -9,7 +9,7 @@
 	url = https://github.com/aantron/websocketaf.git
 [submodule "src/vendor/h2"]
 	path = src/vendor/h2
-	url = https://github.com/aantron/ocaml-h2.git
+	url = https://github.com/novemberkilo/ocaml-h2.git
 [submodule "src/vendor/paf"]
 	path = src/vendor/paf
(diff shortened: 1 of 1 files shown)


ulrikstrid:
Upstream h2 builds and runs on OCaml5, we're using it in multiple projects via Piaf

aantron:
Indeed, thank you!

I missed this PR because I started with local build failures. But while addressing those, I saw @novemberkilo and @patricoferris patch, which, I understand, was live in the OCaml 5.0.0 beta ports opam repo (thank you!). I then noticed that upstream h2 has a similar patch by now, so I decided to point to it.

I've also rebased the renaming branch that changes `H2` to `Dream_h2`, etc. I will upload that shortly.

PR #249, opened 2023-03-07
Cherry-picking Mirage updates

aantron:
@TheLortex This is an FYI PR to track my reviewing and cherry-picking your fork (thank you!) back into the main Dream repo. I've also set up a minimal Mirage dev enivornment so that I can do some actual maintenance. In the past, I fully relied on @dinosaure to do everything (thanks!) and did not even try the most basic Mirage builds due to what I considered a lack of time. I am now building at least for `mirage-unix` as the most minimal, basic check.

After this PR, I will probably set up a CI build to continuously test that the Mirage version is not falling too much behind. I'll probably also replace the current Mirage Let's Encrypt example with something much simpler, a Mirage+Dream Hello World, and resurrect the current Mirage example later, as something for more advanced users. I use the examples myself for simple checks.

Note: this is towards #4.


aantron:
I also plan to point Dream `master` to the renamed forks of the vendored HTTP stack (cc @clecat), and I'll maintain the renamings until we can get the http/af released somehow. The releases of Dream will contain these renamed forks and will no longer conflict with the opam packages. cc @anmonteiro

dream-mirage.opam
```diff
@@ -60,6 +60,7 @@ depends: [
   "tls-mirage"
   "lwt_ppx" {>= "1.2.2"}
   "letsencrypt" {>= "0.3.0"}
+  "emile" {>= "1.1"}
```
dinosaure, line 63: It's a bit unfortunate 😕 . Currently, only `paf-le` (which is another package to handle Let's encrypt certificates) depends on `emile`. Not sure which direction you want to take, but if you don't use `paf-le` and Let's encrypt certificate, you should not have such dependency.

aantron: Thanks! I'll note that for a separate issue. I want to get these commits over into Dream in about the same form that they are already in, and then edit them further.

aantron:
I believe this includes all the commits that could be applied. I skipped some commits related to updating h2, Caqti, etc., because I did that separately in Dream `master` before this PR. The Mirage build in `master` is currently still broken as of this PR, with

```
File "duniverse/dream/src/vendor/paf/lib/alpn.ml", line 231, characters 37-41:
231 |         H2.Client_connection.request conn request ~error_handler
                                           ^^^^
Error: This expression has type unit -> H2.Client_connection.t
       but an expression was expected of type H2.Client_connection.t
       Hint: Did you forget to provide `()' as argument?
```

But that's probably because I advanced h2 farther than this PR advances paf (haven't checked), and I'll fix that after the PR.

@TheLortex, is this still relevant? https://github.com/clecat/dream/commit/eafbbfe61e466c84beb765dee65511916d334ed0

PR #250, opened 2023-03-10
Wrap vendored modules to avoid build conflicts

### LINKED ISSUE #198: Prefix the vendored libraries

aantron:
This FYI PR renames the vendored Gluten, http/af, h2, websocket/af, to `Dream_gluten`, etc., to avoid conflicts with versions of these modules installed from opam.

I took a different, much less invasive and much more maintainable approach to the renaming than in #200. I used that we replace all the `dune` files of the vendored modules with `src/vendor/dune` to wrap e.g. `Gluten` as `Dream_gluten.Gluten`, and then added module aliases at the top of each file that used `Gluten`, so `module Gluten = Dream_gluten.Gluten`. This is in contrast to #200, which renamed the `Gluten` to `Dream_gluten` throughout the code in all the files, and renamed the files. I will check that this truly doesn't conflict with parallel installations of the "real" `Gluten`.

I still have paf to rename afterwards, but it should go the same way as the other four libraries.

cc @clecat @anmonteiro @dinosaure @hannesm @TheLortex

Fixes #198.
Closes #200.

### DIFF
--- a/src/http/adapt.ml
+++ b/src/http/adapt.ml
@@ -5,6 +5,9 @@
 
 
 
+module Httpaf = Dream_httpaf_.Httpaf
+module H2 = Dream_h2.H2
+
 module Stream = Dream_pure.Stream
 module Message = Dream_pure.Message
 
--- a/src/http/dune
+++ b/src/http/dune
@@ -8,17 +8,17 @@
   dream-pure
   dream.server
   dream-httpaf
-  dream-httpaf.gluten
-  dream-httpaf.gluten-lwt-unix
-  dream-httpaf.h2
-  dream-httpaf.h2-lwt-unix
-  dream-httpaf.httpaf
-  dream-httpaf.httpaf-lwt-unix
+  dream-httpaf.dream-gluten
+  dream-httpaf.dream-gluten-lwt-unix
+  dream-httpaf.dream-h2
+  dream-httpaf.dream-h2-lwt-unix
+  dream-httpaf.dream-httpaf_
+  dream-httpaf.dream-httpaf_-lwt-unix
   lwt
   lwt.unix
   lwt_ssl
   ssl
-  dream-httpaf.websocketaf
+  dream-httpaf.dream-websocketaf
  )
  (preprocess (pps lwt_ppx))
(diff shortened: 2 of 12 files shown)


dinosaure:
It's good for me as far as it fixes the initial issue explained by #200 👍 .

hannesm:
Thanks, this looks like a reasonable path forwards. I'm curious why gluten, h2, and paf are still vendored -- can't you just use the latest releases thereof? Removing vendored projects should ease the maintenance.

For http/af I understand it hasn't seen a release in quite some while (and there's still https://github.com/inhabitedtype/httpaf/pull/203 open). websocket/af don't know much about.

Thanks for your work on dream and picking up the outstanding issues.

aantron:
> I'm curious why gluten, h2, and paf are still vendored

h2 [depends](https://github.com/anmonteiro/ocaml-h2/blob/d80e74b8ad78e782dd20eea433517b68f6d3af8d/h2.opam#L24) on http/af. An opam h2 will build against an opam http/af, but we need it to build against Dream's http/af. The same is [true](https://github.com/dinosaure/paf-le-chien/blob/1b22a505e28ba99f4db335f87734d93608c51d89/paf.opam#L29) for paf.

Gluten indeed can be used from opam, and before I settled on this approach I started a branch which unvendored it. But because everything else depends on http/af, it's the only library that can be taken out, and I decided that it's not worth it for the time being. In the past, Gluten was released less frequently and it was convenient (actually, necessary), to be able to point it to any commit. It's best to just keep this benefit for now, given that Gluten is directly related to all the other vendored libraries, and the very incremental benefit of unvendoring only Gluten.

PR #260, opened 2023-04-21
Add dependency on rresult.

tmcgilchrist:
Missing dependency on `rresult` in the `dream-mirage` library.

https://github.com/aantron/dream/blob/master/src/mirage/mirage.ml#L20


aantron:
Thank you!

Issue #233, opened 2022-06-28
Setting an header of empty strings (~headers:[ "", "" ]) causes hang


pckilgore:
`dream~alpha4`, specifically Dream.json.

The workaround is laughable (don't do that), so I didn't spend any time digging to see if it affects other helpers etc.

One possible clue is that if I `dune exec /.app.exe` locally on macos 12.4 (arm64) I don't see any problems, only when built + deployed on `alpine:3.12` (x86).

If I figure out anything more I'll update, but *this is super low priority just want this to be searchable*.

Thanks for Dream!

dangdennis:
taking a look

dangdennis:
i'll look to create a separate repo and link here for reproduction

aantron:
@dangdennis Are you able to test locally in just a regular Dream build, without Docker? Or is Docker required for OCaml on M1?

aantron:
The output during container builds does look like additional other issues worth looking at separately, however.

garyrob:
> @dangdennis Are you able to test locally in just a regular Dream build, without Docker? Or is Docker required for OCaml on M1?

For what it's worth, I'm using an M1 Mac with OCaml without docker and without any problems, but I run iTerm under Rosetta so my development environment is emulating Intel. Haven't done development at all for straight ARM.

dangdennis:
All fine with local opam and dune, no docker.

dangdennis:
@pckilgore did you not run into any ssl dep issues related to alpine?

dangdennis:
Two options from what I can see

1. We filter our `""` headers at each of the response helpers, like at Dream.json [line link](https://github.com/aantron/dream/blob/baa222bb8ec3ac25333d9967341fc4b8245fa77b/src/server/helpers.ml#L75)
2. Add a middleware that always runs that cleans the headers to the middleware chain. I assume it's the [`built_in_middleware`](https://github.com/aantron/dream/blob/baa222bb8ec3ac25333d9967341fc4b8245fa77b/src/http/http.ml#L385)

2 seems correct, and leaves the functionality all optional via `~builtins`.

pckilgore:
> @pckilgore did you not run into any ssl dep issues related to alpine?

```
FROM node:16-alpine AS node
FROM ocaml/opam:alpine as build

# Grab Node & Friends
COPY --from=node /usr/lib /usr/lib
COPY --from=node /usr/local/share /usr/local/share
COPY --from=node /usr/local/lib /usr/local/lib
[...]

PR #262, opened 2023-04-21
fix: middleware to warn if response headers are invalid

dangdennis:
Resolves https://github.com/aantron/dream/issues/233. I was unable to replicate the socket hanging issue on alpine because I couldn't build it due to ssl dep issues. However, I ran the debian build on fly.io, and fly rejects http messages with invalid headers. So we should fix this regardless of linux distro. We can fix the alpine ssl issue separately.

- [x]  Learned that dropping empty headers via the builtin middleware will not work because it runs before the user's middleware, and definitely before the helper methods are called
- [x] Spiral in dune usage lol
- [x] Clean headers at lowest root response call


aantron:
Thanks!

> * Learned that dropping empty headers via the builtin middleware will not work because it runs before the user's middleware, and definitely before the helper methods are called

Not yet a full review, but AFAIK built-in middleware runs both before and after the user's middleware:

```ocaml
let my_middleware user's_middleware = fun request ->
  do_stuff_before;
  let%lwt response = user's_middleware request in
  do_stuff_after;
  response
```

dangdennis:
> Thanks!
> 
> > * Learned that dropping empty headers via the builtin middleware will not work because it runs before the user's middleware, and definitely before the helper methods are called
> 
> Not yet a full review, but AFAIK built-in middleware runs both before and after the user's middleware:
> 
> ```ocaml
> let my_middleware user's_middleware = fun request ->
>   do_stuff_before;
>   let%lwt response = user's_middleware request in
>   do_stuff_after;
>   response
> ```

Ah thanks for clarifying. I'll re-implement this as a built-in middleware built-in then to log a warning if there are empty header keys.

dangdennis:
😆  

> AFAIK built-in middleware runs 

This cracks me up because I'm assuming you've written the middleware logic yourself.

aantron:
> This cracks me up because I'm assuming you've written the middleware logic yourself.

Need to be careful in case I forgot or something has been broken since :)

dangdennis:
unfortunately can't use expect-test against log outputs because we can't control the time when the logger executes. not that i can see any option. since i'm not performing any mutations anyways, think we're safe to not have a test.

test/expect/pure/message/message.ml
```diff
@@ -3,9 +3,7 @@
 
    Copyright 2022 Anton Bachin *)
 
-
-
-let%expect_test _ =
+let%expect_test "middleware runs sequentially onion-style" =
```
dangdennis, line 6: autoformatter kicked in for the file too.

aantron:
I can merge this now to avoid extending this PR unnecessarily. However, as I recall, I wanted to move Dream away from the built-in middleware model. That's why there was only one left, and the plan was to find a way to get rid of it as a built-in middleware as well. I think this warning can go unconditionally into the HTTP adapters -- hopefully in a place where it will be triggered for both HTTP1 and HTTP2. This area:

https://github.com/aantron/dream/blob/a2961cd7fa150903db758fa3737213cddd7ddb3b/src/http/http.ml#L407-L448

already has several wrappers around the user's handler, so we just need to add the middleware here unconditionally. It should run outside of the `catch` middleware, since `catch` might also somehow set an empty header.

dangdennis:
I'd prefer do this as a follow-up too - re: move the header check to h1/h2 paths. I thought it was a little strange to have important checks be opt-in, so I'm glad to see this direction change. The gold standard imo is asp.net & phoenix. Baking in security and best-practices seems best.

test/expect/http/dune
```diff
@@ -0,0 +1,11 @@
...
+ (name test_expect_http)
+ (libraries
+  base
+  dream
+  dream.http
+  lwt
+  lwt.unix
+  ppx_expect.common
+  ppx_inline_test.config
+  ppx_expect.config_types))
```
dangdennis, line 11: leaving these dune files here for the future instead of having to set them up again.

src/http/http.ml
```diff
@@ -380,15 +380,20 @@ let ocaml_tls = {
...
+  let invalid_headers_exist = Message.all_headers response |> List.exists (fun (key, _) -> String.trim key = "") in
+  if invalid_headers_exist then
+    log.warning (fun log ->
+      log "A response header is empty or contains only whitespace. This is not allowed by RFC 7230.");
+  Lwt.return response
 
 let built_in_middleware error_handler =
   Message.pipeline [
     Catch.catch (Error_handler.app error_handler);
+    check_headers_middleware;
```
aantron, line 394: I think this should be the outer middleware, as the error handler can also be a source of empty headers, while, conversely, `check_headers_middleware` cannot raise any exceptions.

aantron:
> I'd prefer do this as a follow-up too - re: move the header check to h1/h2 paths. I thought it was a little strange to have important checks be opt-in, so I'm glad to see this direction change. The gold standard imo is asp.net & phoenix. Baking in security and best-practices seems best.

Not sure what you mean here -- built-in middlewares are/were opt-out, they are included in the middleware stack by default.

dangdennis:
> > I'd prefer do this as a follow-up too - re: move the header check to h1/h2 paths. I thought it was a little strange to have important checks be opt-in, so I'm glad to see this direction change. The gold standard imo is asp.net & phoenix. Baking in security and best-practices seems best.
> 
> Not sure what you mean here -- built-in middlewares are/were opt-out, they are included in the middleware stack by default.

Oops! Yes you're right. It's already defaulted to true. I forgot to check the actual default value 😅 . I only ever read the lines below, and not specifically the args set on `run`.

```
  let user's_dream_handler =
    if builtins then
      built_in_middleware error_handler user's_dream_handler
    else
      user's_dream_handler
  in
```

aantron:
I've made a few edits. Notably:

```diff
diff --git a/src/http/http.ml b/src/http/http.ml
index 11943be..d1b7614 100644
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -390,7 +390,7 @@ let check_headers_middleware next_handler request =
   in
   if invalid_headers_exist then
     log.warning (fun log ->
-      log "A response header is empty or contains only whitespace");
+      log ~request "A response header is empty or contains only whitespace");
   Lwt.return response

 let built_in_middleware error_handler =
```

changes the log output from

```
23.04.23 14:13:42.184      dream.http  WARN A response header is empty or contains only whitespace
```

to

```
23.04.23 14:13:42.184      dream.http  WARN REQ 1 A response header is empty or contains only whitespace
```

I then tested example `9-error` to see if this PR would warn about empty headers set in the error handler:

```diff
diff --git a/example/9-error/error.eml.ml b/example/9-error/error.eml.ml
index 515091c..8a995d7 100644
--- a/example/9-error/error.eml.ml
+++ b/example/9-error/error.eml.ml
@@ -12,6 +12,7 @@ let my_error_template _error debug_info suggested_response =
     </body>
     </html>
   end;
+  Dream.set_header suggested_response "" "abc";
   Lwt.return suggested_response

 let () =
```

The PR didn't warn about this header, as expected, so I added

```diff
diff --git a/src/http/http.ml b/src/http/http.ml
index d1b7614..2deb1db 100644
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -395,8 +395,8 @@ let check_headers_middleware next_handler request =

 let built_in_middleware error_handler =
   Message.pipeline [
-    Catch.catch (Error_handler.app error_handler);
     check_headers_middleware;
+    Catch.catch (Error_handler.app error_handler);
   ]


```

It now warns about empty headers set in error handlers:

```
23.04.23 14:17:19.285       dream.log  INFO REQ 1 GET / 127.0.0.1:39224 Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/112.0.0.0 Safari/537.36
23.04.23 14:17:19.285       dream.log  WARN REQ 1 404 in 62 μs
23.04.23 14:17:19.285      dream.http  WARN REQ 1 A response header is empty or contains only whitespace
```

aantron:
Thank you! As for built-in middleware, it's probably fine to leave the implementation in `http.ml` as it is now. I'll probably just hide them (their remainder, `Dream.catch`) from `dream.mli`. All the other former built-in middlewares have already been factored away anyway, and all that's left is a few that we want to keep around unconditionally, and it's fine to internally interpret them as a middleware stack.

PR #263, opened 2023-04-21
fix: update ocamlformat version so autoformatting actually works

dangdennis:
macos ventura 13.2.1
m1
ocaml 4.14 

I noticed that autoformat doesn't work until I updated the version in the `.ocamlformat` file. Unsure why, but now my vscode actually formats. It's possible that you want the current style, but updating the ocamlformat version adheres to some new defaults of theirs. But I think just getting the autoformat to actually work is a great win.

Following this, it'd be really helpful to have one PR (or a few) format the entire project.


aantron:
Thanks!

PR #265, opened 2023-04-22
Add doc about trailing slashes

### LINKED ISSUE #244: Difference between path and path/ in the router
Say I have the scope:
```ocaml
Dream.scope "/users" []
  [ Dream.get "/" (fun _ -> Dream.html "Hello!") ]
```

When I navigate to `/users` I expect to see `Hello!` but instead I get a 404. If I navigate to `/users/` however, I will see `Hello!`. I don't think this is intentional.

If I want it to catch without the final slash, I have to add another route to the scope:
```ocaml
Dream.scope "/users" []
  [ Dream.get "/" (fun _ -> Dream.html "Hello!")
  ; Dream.get "" (fun _ -> Dream.html "Hello!")
  ]
```
Then it is working as expected.

rawleyfowler:
fixes #244

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -1470,7 +1470,7 @@ val router : route list -> handler
     because, in the future, it may be possible to query routes for site
     structure metadata.
     
-    Dream's router is {{:https://www.rfc-editor.org/rfc/rfc1738#section-3.3} RFC 1738} compliant.
+    Dream's router is {{:https://www.rfc-editor.org/rfc/rfc1738#section-3.3} RFC 1738 Section 3.3} compliant.
     This means routes that end with [/], and routes that don't end with [/] will be treated
     as separate entities. *)
 
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -1468,7 +1468,13 @@ val router : route list -> handler
     include its own router, into a subsite. However, it is better to compose
     sites with routes and {!Dream.scope} rather than opaque handlers and [**],
     because, in the future, it may be possible to query routes for site
-    structure metadata. *)
+    structure metadata.
+    
+    Note, Dream's router differentiates between routes that end with a [/] and routes
+    that don't. For example [/foo/] and [/foo] are treated as different
+    routes. This is to enforce the idea that routes without a final [/] 
+    should serve a file-like response, while routes that do end with [/] offer a
+    directory-like response. *)
 
 val get     : string -> handler -> route
 (** Forwards [`GET] requests for the given path to the handler.
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -1470,8 +1470,8 @@ val router : route list -> handler
     because, in the future, it may be possible to query routes for site
     structure metadata.
     
-    Dream's router is {{:https://www.rfc-editor.org/rfc/rfc1738#section-3.3} RFC 1738} compliant,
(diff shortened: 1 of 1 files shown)


aantron:
@rawleyfowler Thanks! Could you do a brief search for a better resource on this than the [SO thread](https://stackoverflow.com/questions/5948659/when-should-i-use-a-trailing-slash-in-my-url) I linked in #244? Rather than explaining this in Dream's docs, I'd like to briefly mention it in one sentence, and link the interested user to a more extensive discussion somewhere else. Perhaps even one of the HTML [RFCs](https://datatracker.ietf.org/doc/html/rfc7230#section-1) says something directly about this. I'd basically like to warn the user that this difference exists, but let the reasoning be given in detail elsewhere.

aantron:
BTW we can also merge this PR and open a separate issue in case someone else is able to offer a reference later!

rawleyfowler:
@aantron I'll fix it up, sorry it was 2AM when you replied originally :)

rawleyfowler:
@aantron Should be good-to-go, RFC 1738 Section 3.3 has us covered!

aantron:
Thanks. RFC 1738 doesn't seem to really say anything about this issue, neither in section 3.3, nor elsewhere. Section 3.3:

```
   The HTTP URL scheme is used to designate Internet resources
   accessible using HTTP (HyperText Transfer Protocol).

   The HTTP protocol is specified elsewhere. This specification only
   describes the syntax of HTTP URLs.

   An HTTP URL takes the form:

      http://<host>:<port>/<path>?<searchpart>

   where <host> and <port> are as described in [Section 3.1](https://www.rfc-editor.org/rfc/rfc1738#section-3.1). If :<port>
   is omitted, the port defaults to 80.  No user name or password is
   allowed.  <path> is an HTTP selector, and <searchpart> is a query
   string. The <path> is optional, as is the <searchpart> and its
   preceding "?". If neither <path> nor <searchpart> is present, the "/"
   may also be omitted.

   Within the <path> and <searchpart> components, "/", ";", "?" are
   reserved.  The "/" character may be used within HTTP to designate a
   hierarchical structure.
```

So far, the closest was I found was RFC 3986 but it's way too arcane to use as a reference in Dream's docs. There is this [SO answer](https://stackoverflow.com/a/61549137) that mentions it. The answer above that mentions the difference in a different way, and vaguely mentions how relative links are interpreted differently in the two cases. I don't really see a clear reference on the difference yet that mentions all these things in an understandable way.

I think it's fine to merge without it, I just wanted to see if something like that exists online :)

rawleyfowler:
Yeah I looked around, there isn't much. There is a Wikipedia article but I don't know how I feel about linking that, as it doesn't cover a lot either. Maybe we can merge now and at-least have it mentioned then see if something nicer comes around later.

rawleyfowler:
How do we feel about linking the issue in the docs?

aantron:
That might be a good idea. Perhaps the comment I wrote [here](https://github.com/aantron/dream/issues/244#issuecomment-1511624682) is (unsurprisingly) what I am looking for as something to link to, sadly no more detailed such thing seems to exist.

rawleyfowler:
@aantron Hopefully it's good to go now!

aantron:
Many thanks!

Issue #81, opened 2021-06-29
Recommend watchexec for live_reload example


thangngoc89:
I have been trying several solution over the past few days and so far, watchexec (https://github.com/watchexec/watchexec) is the most reliable one and it’s fast. 

fswatch + custom script in the current example keep failing to kill the process, I have to kill it manually several times

tcoopman:
I haven't looked at this, but shouldn't we be able to use `dune --watch` for this?

thangngoc89:
@tcoopman I think the reason have been explained in live-reload example. dune —watch doesn’t kill the server process

tcoopman:
ok, just found the issue on dune: https://github.com/ocaml/dune/issues/2934

aantron:
Also cc @tmattio, in case watchexec is good for the example script in [dream-liverelaod](https://github.com/tmattio/dream-livereload#usage).

aantron:
From https://github.com/aantron/dream/issues/139#issuecomment-884672526, a command line for killing any listening Dream processes (seemingly using port 5000) and restarting:

> ```
> lsof -i tcp:5000 | grep LISTEN | awk '{print $2}' | xargs kill -9 && dune exec ./src/main.exe
> ```

cemerick:
It seems that the dune limitation is confined to `dune exec`. I've had good results by setting up a "test" executable and alias (my dream app needs a number of small tweaks for a dev/test context anyway):

```
(executable (name runapp)
  (libraries supervisor)
  (modules Runapp)
  (flags (:standard -g))
  (preprocess (pps ppx_deriving.std)))

(rule
  (alias runapp)
  (deps runapp.exe)
  (action (run ./runapp.exe)))
```

and then I run it with `dune build @runapp -w --force --no-buffer`. No need for an additional filesystem watcher, explicit process reaping, etc.

aantron:
Does this automatically kill a running `runapp.exe` whenever you make changes to the code?

cemerick:
So far in my experience, yes. I've not used fswatch at all.

dangdennis:
Can close because dune exec -w is now available

aantron:
Thanks! I just tried this out and it seems to kill the running server now -- very nice! We probably need to update the [watch example](https://github.com/aantron/dream/tree/master/example/w-fswatch#files).

tmattio:
[...]

PR #266, opened 2023-04-23
docs: update fswatch example to use dune exec -w

dangdennis:
Since fswatch isn't used anymore, I've renamed the example to `w-watch-mode`. It's now just a nice little tip for anyone getting started with dune and dream. 

~~R esolves https://github.com/aantron/dream/issues/81~~


aantron:
Thank you!

I undid the directory move to help git understand the history. When one combines a massive rewrite (like the README in this PR) together with a renaming, git thinks that a file was deleted and a new one was created. I left the rewriting in this PR, and I'll do the renaming in a separate commit after this PR, and also update all the links to the example from the README and the example index.

dangdennis:
Awesome. I’ll do the git mv manually next time then. 😂 high bar for git cleanliness here.

dangdennis:
great catch on the links.

aantron:
> Awesome. I’ll do the git mv manually next time then. 😂 high bar for git cleanliness here.

Yeah I too often have to dig in git history and the cleaner it is the easier it is in the future, git history is like a form of documentation, especially when you eventually are able to reach the various PRs that caused something. There is often a lack of comments in the code but plenty of discussion in the PR. So it helps the archaelogy to keep the history clear :)

PR #279, opened 2023-05-13
Add a `Dream.livereload` middleware

tmattio:
The code is adapted from [dream-livereload](https://github.com/tmattio/dream-livereload), itself adapted from the w-live-reload example.
We update the w-live-reload to use the newly introduced middleware.

### DIFF
--- a/example/w-live-reload/README.md
+++ b/example/w-live-reload/README.md
@@ -2,83 +2,20 @@
 
 <br>
 
-This example shows a simple live reloading setup. It works by injecting a script
-into the `<head>` of HTML documents. The script opens a WebSocket back to the
-server. If the WebSocket gets closed, the script tries to reconnect. When the
-server comes back up, the client is able to reconnect, and reloads itself.
+This example shows a simple live reloading setup using the `Dream.livereload`
+middleware. It works by injecting a script into the `<head>` of HTML documents.
+The script opens a WebSocket back to the server. If the WebSocket gets closed,
+the script tries to reconnect. When the server comes back up, the client is able
+to reconnect and reloads itself.
 
-```js
-var socketUrl = "ws://" + location.host + "/_live-reload"
-var socket = new WebSocket(socketUrl);
-
-socket.onclose = function(event) {
-  const intervalMs = 100;
-  const attempts = 100;
-  let attempt = 0;
-
-  function reload() {
-    ++attempt;
-
-    if(attempt > attempts) {
-      console.error("Could not reconnect to server");
-      return;
-    }
-
-    reconnectSocket = new WebSocket(socketUrl);
-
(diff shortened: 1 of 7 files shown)


aantron:
@tmattio, I think we should remove the `unit` argument. The `unit` argument makes `Dream.livereload ()` compose better with `Dream.no_middleware` inside an `if`-expression. But we already have middlewares that may be used in development but that don't have the `unit` argument and therefore don't compose well inside `if`, for example, `Dream.memory_sessions`:

https://github.com/aantron/dream/blob/6e666d4c7079b29d2f287febe312cae20c116900/src/dream.mli#L1573

Also, this approach is not future-proof, because any middleware could gain an optional argument in the future, and we won't be able to add `unit` arguments to them without breaking users. This does mean that adding any optionals to any middlewares will break them if they are used inside `if`-expressions, but that's just a general flaw of optional arguments in OCaml.

We will just add an `Dream.if_` [combinator](https://aantron.github.io/dream/#middleware) later if needed. For example:

```ocaml
let if_ : bool -> then_:Dream.middleware -> Dream.middleware =
    fun condition ~then_ ->
  if condition then
    then_ 
  else
    Dream.no_middleware
```

It's also possible for the user to work around this using a type constraint, `Dream.livereload : middleware`, although, of course, this is awkward.

Separately, are the `?script` and `?path` options really needed? We can probably get by for now without them. I'd rather add them later on request, than support them when there is no evidence of them being used (I mean, I don't have any -- you may well have!).

example/w-live-reload/README.md
```diff
@@ -2,83 +2,20 @@
...
 
-This example shows a simple live reloading setup. It works by injecting a script
-into the `<head>` of HTML documents. The script opens a WebSocket back to the
-server. If the WebSocket gets closed, the script tries to reconnect. When the
-server comes back up, the client is able to reconnect, and reloads itself.
+This example shows a simple live reloading setup using the `Dream.livereload`
+middleware. It works by injecting a script into the `<head>` of HTML documents.
+The script opens a WebSocket back to the server. If the WebSocket gets closed,
+the script tries to reconnect. When the server comes back up, the client is able
+to reconnect and reloads itself.
```
aantron, line 9: This example should show how to disable the middleware conditionally for production.

src/server/livereload.ml
```diff
@@ -0,0 +1,78 @@
+(* This file is part of Dream, released under the MIT license. See LICENSE.md
+   for details, or visit https://github.com/aantron/dream.
+
+   Copyright 2021 Anton Bachin *)
```
aantron, line 4: @tmattio You should credit yourself :)

aantron:
@tmattio, I'm taking over this PR to merge it in!

PR #280, opened 2023-05-13
Add encoding primitives and a compression middleware

tmattio:
This essentially upstreams [dream-encoding](https://github.com/tmattio/dream-encoding).

It adds a new dependency on [`decompress`](https://github.com/mirage/decompress).

Two new middleware are added:

- Dream.compress
- Dream.decompress

And we add a new Encoding section to the API with the following functions:

- Dream.with_encoded_body
- Dream.accepted_encodings
- Dream.accepted_encodings_with_weights
- Dream.content_encodings
- Dream.preferred_content_encoding

I'm happy to keep this in a separate package, I'm opening a PR in case you're interested in having this upstream, but don't hesitate to close the PR if that's outside of the scope of first-class features you see.


src/unix/encoding.ml
```diff
@@ -0,0 +1,201 @@
+(* This file is part of Dream, released under the MIT license. See LICENSE.md
```
aantron, line 1: Why is this file in `src/unix/`? Can't we do compression/decompression "purely," i.e. in memory?

src/unix/encoding.ml
```diff
@@ -0,0 +1,201 @@
...
+let compress handler req =
+  let%lwt response = handler req in
+  let preferred_algorithm = preferred_content_encoding req in
+  match preferred_algorithm with
+  | None -> Lwt.return response
+  | Some algorithm ->
+    log.info (fun log ->
+        log ~request:req "Compressing request with algorithm: %s"
+          (algorithm_to_string algorithm));
+    let%lwt body = Message.body response in
```
aantron, line 180: Note: it should be possible to have a streaming compressor/decompressor with the current Dream API. I will probably write an example that does a streaming transformation of a body stream in some way, and a compressor/decompressor would be built analogously.

src/unix/encoding.ml
```diff
@@ -0,0 +1,201 @@
...
+    | [] -> Ok content
+    | (`Deflate as el) :: rest | (`Gzip as el) :: rest ->
+      Result.bind (inflate_string ~algorithm:el content) (aux rest)
+    | _ :: _rest -> Error (`Msg "Unsopported encoding directive")
+  in
+  let algorithms = content_encodings req in
+  match algorithms with
+  | None -> handler req
+  | Some algorithms -> (
+    let%lwt body = Message.body req in
```
aantron, line 195: (Same note here about streaming)

aantron (commented):
Thanks! I see an argument for not having this in Dream, which is this, or variants of this, are also needed on the client side (i.e., in Hyper).

PR #285, opened 2023-06-22
Autodetecting default theme

unorsk:
Not everyone uses "dark" mode by default, so I thought it would be a nice change if the docs site could figure out the system theme and use it, otherwise fall back to "dark".

Then I also added an event listener that would change theme once the system theme changes, unless there is a preferred theme in the local storage.

### DIFF
--- a/docs/web/site/docs.js
+++ b/docs/web/site/docs.js
@@ -52,11 +52,13 @@ window.onscroll = scroll;
 
 /* Theme mode */
 
-var THEME_MODE_KEY = "dream-theme"
+var THEME_MODE_KEY = "dream-theme";
+var DARK_THEME = "dark";
+var LIGHT_THEME = "light";
 
 function apply_theme(theme) {
-  if (theme === "light") {
-    document.body.setAttribute("data-theme", "light");
+  if (theme === LIGHT_THEME) {
+    document.body.setAttribute("data-theme", LIGHT_THEME);
   } else {
     document.body.removeAttribute("data-theme");
   }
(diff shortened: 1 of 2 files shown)


aantron:
@unorsk, thank you! I personally tend to lean towards dark theme always, even when my browser and/or OS are set to light. At least in the browers I use, I personally tend to distinguish between my UI theme preferences and my website default theme preferences. And at least Brave doesn't offer different settings for each -- I would be happy to respect a user's preference for default website colors. So the current behavior reflects my (quite subjective) preferences. We considered (on stream, and IIRC, in the original PR) having the the theme respect the browser theme setting, but ended up not doing it.

Please let me know if there is some kind of established best practice on this, and we are violating it by not having the docs site respect the browser theme.

unorsk:
> @unorsk, thank you! I personally tend to lean towards dark theme always, even when my browser and/or OS are set to light. At least in the browers I use, I personally tend to distinguish between my UI theme preferences and my website default theme preferences. And at least Brave doesn't offer different settings for each -- I would be happy to respect a user's preference for default website colors. So the current behavior reflects my (quite subjective) preferences. We considered (on stream, and IIRC, in the original PR) having the the theme respect the browser theme setting, but ended up not doing it.
> 
> Please let me know if there is some kind of established best practice on this, and we are violating it by not having the docs site respect the browser theme.

Hey!

I personally use light themes for almost everything 😅 And every time I stumble upon a website with a dark theme it kinda hurts and it takes time for the eyes to get adjusted. But then when I see the option to change the colors I am like "why wouldn't you just detect my system theme and use it instead?" Hence the PR :)

PR #291, opened 2023-07-24
chore(dream): re-export dream-pure library

glennsl:
Using the template preprocessor  assumes `dream-pure` is re-exported, which it won't be if `implicit_transitive_deps` is disabled in the consuming project, unless the library is explicitly re-exported. This does that.

For more information:
- [Documentation for `re_export`](https://dune.readthedocs.io/en/stable/reference/library-dependencies.html#re-exported-dependencies)
- [Documentation for `implicit_transitive_deps`](https://dune.readthedocs.io/en/stable/dune-files.html#implicit-transitive-deps)
- [An example of someone likely tripping over this on Stack Overflow](https://stackoverflow.com/q/76750178/7943564)

### DIFF
--- a/src/dune
+++ b/src/dune
@@ -9,7 +9,7 @@
   dream.http
   dream.server
   dream.unix
-  dream-pure
+  (re_export dream-pure)
   dream.sql
   fmt.tty
   graphql-lwt


aantron:
Thanks! This was actually an oversight. The templater should have been emitting `Dream.html_escape`. Probably left over from some refactoring. The templater uses the `Dream` module directly in other code it emits. I tested the templater against a version of example `7-template` with `(implicit_transitive_deps false)` and it now works.

samoht:
I think this breaks MirageOS support, as now you need to depends on `dream` when using `eml` files

aantron:
@samoht As I understand it, `dream-mirage` already [depends](https://github.com/aantron/dream/blob/78c79e3328414768fa2ede0623d741f63ea5c4f1/dream-mirage.opam#L53) on `dream`. Do you mean something extra?

samoht:
It doesn't depend on `dream` in the [dune file](https://github.com/aantron/dream/blob/78c79e3328414768fa2ede0623d741f63ea5c4f1/src/mirage/dune)

aantron:
@hannesm, @dinosaure, could you comment on whether this indeed breaks Mirage? I'm ready to revert 99388503ccd69272f596353f8d346f807858b40f and merge in the original commit here, if needed.

Issue #275, opened 2023-05-11
Issues with Docker builds in examples


aantron:
From https://github.com/aantron/dream/issues/233#issuecomment-1516405747:

> Testing with `z-fly` and `z-docker-opam` examples, and I'm encountering build issues within just the docker compose up process even before running a request with the empty key-value header.
> 
> ```
> (* app.ml *)
> let () =
>   Dream.run ~interface:"0.0.0.0"
>   @@ Dream.logger
>   @@ Dream.router [
>     Dream.get "/" (fun _ -> Dream.html ~headers:[("", "")] "Dream deployed on Fly.io!");
>   ]
> ```
> 
> Local tests first on an m1 mac:
> 
> `z-docker-opam` local error with `docker compose up`
> 
> ```
> z-docker-opam-web-1  | Error loading shared library libssl.so.3: No such file or directory (needed by /bin/app)
> z-docker-opam-web-1  | Error loading shared library libcrypto.so.3: No such file or directory (needed by /bin/app)
> z-docker-opam-web-1  | Error relocating /bin/app: SSL_CTX_get_cert_store: symbol not found
> z-docker-opam-web-1  | Error relocating /bin/app: SSL_load_client_CA_file: symbol not found
> ...I truncated because we get the picture
> ```
> 
> `z-fly` local error with `docker compose up`
> 
> ```
>  > [build  5/12] RUN npm install esy:
> #9 15.35 npm ERR! code 1
> #9 15.35 npm ERR! path /build/node_modules/esy
> #9 15.35 npm ERR! command failed
> #9 15.35 npm ERR! command sh -c node -e "process.env['OCAML_VERSION'] = process.platform == 'linux' ? '4.12.0-musl.static.flambda': '4.12.0'; process.env['OCAML_PKG_NAME'] = 'ocaml'; require('./postinstall.js')"
> #9 15.35 npm ERR! internal/fs/utils.js:269
> #9 15.35 npm ERR!     throw err;
> #9 15.35 npm ERR!     ^
> #9 15.35 npm ERR! 
> #9 15.35 npm ERR! Error: ENOENT: no such file or directory, scandir '/build/node_modules/esy/platform-linux-arm64/bin'
> ```
> 
> To test a remote build, I've deployed both examples to fly.io.
> 
> `z-docker-opam` fails with the same missing missing ssl files
> 
[...]

PR #300, opened 2023-10-06
example/z-docker-opam: update alpine version

pedrofgd:
I've encountered this issue #275 related to `z-docker-opam` example when first tried to run it locally.

I saw this `Error loading shared library libssl.so.3: No such file or directory (needed by /bin/app)` on Mac M1, also on Windows and WSL with Ubuntu.

Update the alpine version from `3.12` to the most recent `3.18.4` had solved things here on all 3 environments (removing the `LOGGING_DRIVER` variable from `deploy.sh`).

### DIFF
--- a/example/z-docker-opam/Dockerfile
+++ b/example/z-docker-opam/Dockerfile
@@ -15,7 +15,7 @@ RUN opam exec -- dune build
 
 
 
-FROM alpine:3.12 as run
+FROM alpine:3.18.4 as run
 
 RUN apk add --update libev
 


pedrofgd:
I've also read a part of a discussion on using Debian instead of Alpine in this example: https://github.com/aantron/dream/pull/57#discussion_r627325258

`ocaml/opam` now has this [debian-12-ocaml-4.08](https://hub.docker.com/layers/ocaml/opam/debian-12-ocaml-4.08/images/sha256-0a2b6f62ba1b86c142112592be0590cc59dbd74bf557562cd68619370f19d39d?context=explore) Docker image. I could try to work on that change if it's appropriate.

aantron:
Many thanks! I think we should stick with Alpine for now, but if we do get more observations/complaints about unpatched CVEs, etc., we should probably move away from it then.

PR #302, opened 2023-10-12
Upgrade Caqti to 2.0.1

paurkedal:
Caqti 2 includes some breaking changes, so I'm proposing this upgrade. To run the test suite I had comment out `Mirage.form_tag` and disable the z-playground example, which is of course not included in this PR.

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -50,8 +50,8 @@ depends: [
   "base-unix"
   "bigarray-compat"
   "camlp-streams"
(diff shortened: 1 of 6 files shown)


aantron:
Thank you! Any strong reason why the lower bound is 2.0.1 instead of 2.0.0?

paurkedal:
Well, either version will work, since 2.0.0 was abandoned and never published in opam.

aantron:
Thank you!

PR #311, opened 2023-12-24
Small fix on examples/r-template-files/

renatillas:
This code example was wrong as the Dream.param parameters were in the wrong order.


aantron:
The README got out of sync with the code in the example. Thank you for fixing it!

PR #312, opened 2023-12-26
Update docker install command

renatillas:
Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).


aantron:
Thank you!

PR #314, opened 2024-01-01
Fix typo in doc of Dream.router

acieroid:
This fixes a small typo in the documentation of Dream.router.
Dream.param takes the request as first argument, then the request, whereas the doc has reversed the arguments


aantron:
Thank you!

PR #326, opened 2024-06-08
Add html_of_jsx template

davesnx:
This might be broken, since current opam release expects > 5.0.0, but will release a new package (0.0.4) without the lower band on 5.


aantron:
Thank you! This looks amazing (as does the sibling mlx PR by @andreypopp).

I was trying this out with html_of_jsx 0.0.4 and Reason 3.11.0 installed, and got

```
File "dream/example/r-html_of_jsx/html_of_jsx.re", line 15, characters 41-42:
15 |       (_ => Dream.html(Jsx.render(<greet ~who="world" />))))),
                                              ^
Error: Syntax error
```

Do you know what's causing that?

davesnx:
Fixed the syntax, indeed is similar to mlx but with Reason syntax

aantron:
Thanks! I had to make [these changes](https://github.com/aantron/dream/pull/326/commits/de0a86d304a15e149971a11602f2b5a65e6ab41c) to get the example to compile. Could you take a quick look, just in case I am using something outdated? Would be very helpful, as I have no recent experience with JSX in Reason.

aantron:
Thank you!

PR #327, opened 2024-07-03
Port to Eio and Cohttp

aantron:
This is a messy draft port of Dream to [Eio](https://github.com/ocaml-multicore/eio) using [cohttp-eio](https://github.com/mirage/ocaml-cohttp). On HTTP/1, it works better than the previous port attempts (#254, #194) and should be more maintainable, including because cohttp-eio offers a direct-style interface (the advantages of which this port still needs to propagate further into the Dream code), but a lot of features will be missing even after the port is complete -- HTTP/2 support, GraphQL, and others, because the corresponding libraries are not ported to Eio or not available with Cohttp. A lot of features, like Caqti integration, that will be available when this port is done, are temporarily disabled for now, and will be added back as the port progresses. WebSockets will require additional work.

The current status of the port is that I am working through various deadlocks and corner cases of using Eio "promises" together with fibers, which are caused by the existing code of Dream assuming a sort of promise-callback soup, where any callback can be called on the one big stack under `Lwt.main` and make progress, while with fibers and Eio blocking the wrong fiber to wait on a promise can block a stack that would have been used to make progress to resolve that promise. The semantics are therefore substantially different. This was a major problem with the approach in #194 and #254. It is substantially negated by thoroughly using direct style wherever possible, rather than promises or callbacks.


smorimoto:
6.0 just shipped! https://github.com/mirage/ocaml-cohttp/releases/tag/v6.0.0

src/http/driver.ml
```diff
@@ -456,34 +556,52 @@ let serve_with_details
...
       match addresses with
       | [] ->
         Printf.ksprintf failwith "Dream.%s: no interface with address %s"
           caller_function_for_error_messages interface
       | address::_ ->
-        Lwt.return Lwt_unix.(address.ai_addr)
+        address
   in
 
+  (* TODO Value of the backlog argument? *)
```
yawaramin, line 579: Seems like backlog should be around 128: https://stackoverflow.com/a/36452474/20371

src/pure/message.ml
```diff
@@ -198,15 +197,16 @@ let body message =
...
 
 let set_body message body =
-  message.body <- Some (Lwt.return body);
+  message.body <- Some (Eio.Promise.create_resolved (Ok body));
   match message.kind with
   | Request -> message.server_stream <- Stream.string body
   | Response -> message.client_stream <- Stream.string body
 
 let set_content_length_headers message =
+  (* TODO Restore.
```
yawaramin, line 209: Seems like we can use `Eio.Promise.peek` to do something similar to `Lwt.poll` here?

src/pure/stream.mli
```diff
@@ -15,8 +15,6 @@ type stream
 
 type buffer =
   (char, Bigarray.int8_unsigned_elt, Bigarray.c_layout) Bigarray.Array1.t
-type 'a promise =
```
yawaramin, line 18: Ironically, we could probably do `type 'a promise = 'a Eio.Promise.t` here.

src/server/log.ml
```diff
@@ -70,8 +70,10 @@ let logs_lib_tag : string Logs.Tag.def =
 
 (* Lwt sequence-associated storage key used to pass request ids for use when
    ~request is not provided. *)
+(* TODO Is there an equivalent mechanism with Eio?
```
yawaramin, line 73: Looks like yes: https://ocaml-multicore.github.io/eio/eio/Eio/Fiber/index.html#fiber-local-variables

> Fiber-local variables are particularly useful for attaching extra information for debugging, such as a request ID that the log system can include in all logged messages.

src/server/log.ml
```diff
@@ -383,6 +388,7 @@ let log =
 
 
 let set_up_exception_hook () =
+  (* TODO Is there an Eio equivalent?
```
yawaramin, line 391: I think no, because to my understanding every fiber in Eio propagates its exceptions upwards to its parent fibers and eventually up to the root fiber.

dream.opam
```diff
@@ -51,28 +51,25 @@ depends: [
...
   "magic-mime"
   "markup" {>= "1.0.2"}
   "mirage-clock" {>= "3.0.0"}  # now_d_ps : unit -> int * int64.
   "mirage-crypto" {>= "0.8.1"}  # AES-256-GCM.
   "mirage-crypto-rng"
-  "mirage-crypto-rng-lwt"
+  "mirage-crypto-rng-eio"
   "multipart_form" {>= "0.4.0"}
-  "multipart_form-lwt"
   "ocaml" {>= "4.08.0"}
```
yawaramin, line 73: I think we can just remove the OCaml version since Eio will force >5

yawaramin:
Trying to see if I can get this branch to run. There's been some bitrot, especially the Mirage folks have gotten rid of cstruct and moved to string/bytes. Hacking and slashing now...

kognise:
I think this is super important, is there any way I can help speed up the work to port dream to eio?

Willenbrink:
Just want to put out there that I have fixed some of the issues in yawaramin's fork which can be found here: [https://github.com/Willenbrink/dream/tree/cohttp-eio](https://github.com/Willenbrink/dream/tree/cohttp-eio).

All the tests pass and the examples we looked at work too. I've also rebased the changes on top of the additional tests I've written in #396 which also pass.

yawaramin:
@Willenbrink great work! Checking out your fork.

PR #330, opened 2024-07-10
Examples: add w-mlx

andreypopp:
Adds an example using [mlx](https://github.com/ocaml-mlx/mlx).

mlx is an OCaml dialect which extends regular OCaml syntax with JSX expressions. The dialect features first class support from merlin/lsp.

### DIFF
--- a/example/w-mlx/README.md
+++ b/example/w-mlx/README.md
@@ -1,0 +1,35 @@
+# `w-mlx`
+
+<br>
+
+[mlx](https://github.com/ocaml-mlx/mlx), an OCaml syntax dialect which adds JSX
+expressions, can be used with Dream for generating HTML. 
+
+```ocaml
+let greet ~who () =
+  <html>
+    <head>
+      <title>"Greeting"</title>
+    </head>
+    <body>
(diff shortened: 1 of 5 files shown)


aantron:
Thanks! This looks great! I'm playing around with it.

What is needed to get the syntax highlighting to work? I have this example building and running with the exact code in this PR. I am using:

VSCode 1.92.1
OCaml Platform 1.20.0
OCaml 5.2.0
Dune 3.16.0
ocaml-lsp-server 1.18.0~5.2preview
ocamlmerlin-mlx 0.9
mlx 0.9

andreypopp:
> What is needed to get the syntax highlighting to work? I have this example building and running with the exact code in this PR. I am using:

Unfortunately this needs an extension for vscode which doesn't yet exists.

Though there exists a [tree-sitter-mlx](https://github.com/ocaml-mlx/tree-sitter-mlx) parser I use in an [ocaml_mlx.nvim](https://github.com/ocaml-mlx/ocaml_mlx.nvim) so if there's some generic vscode extensions which could use tree sitter parsers for highlighting then it'd work now.

aantron:
Thank you!

Issue #174, opened 2021-11-14
signal handling in dockerfile


joprice:
When trying out one of the docker examples like `example/z-docker-opam`, I noticed I can't seem to stop the docker without using `docker kill`. Adding an entrypoint like tini launches the process as pid 1 and makes the signal handling work as expected:

```dockerfile
RUN apk add --update libev gmp tini
...
ENV TINI_VERSION v0.19.0
ADD https://github.com/krallin/tini/releases/download/${TINI_VERSION}/tini-muslc-amd64 /tini
RUN chmod +x /tini
ENTRYPOINT ["/tini", "--"]
CMD ["/bin/app"]
```

(I also had to add `gmp-dev` in the build stage and `gmp` in the run stage. I assume this is due to a dependency change and I can pr this change alone to fix the image.)

ulrikstrid:
About the GMP issue:
I think adding `conf-gmp` as a dependency in esy should solve it without doing anything else. If it's not enough I can find the correct resolution to make it work.

joprice:
For opam, I also found this tip [here](https://medium.com/@bobbypriambodo/lightweight-ocaml-docker-images-with-multi-stage-builds-f7a060c7fce4) for installing any extra deps using `opam depext`:

```dockerfile
# in builder
RUN opam depext | sed 's/-dev//' > depexts`
...
# in run
COPY --from=0 /home/opam/depexts depexts
RUN cat depexts | xargs apk add && rm -rf /var/cache/apk/*
```

aantron:
Thanks!

I won't be able to look into this immediately, but I'll be happy to merge any PRs with improvements, and in the meantime I hope this issue helps others.

yawaramin:
This should be solveable by adding a `~stop` promise argument to the `Dream.run ...` callsite and resolving the promise on SIGTERM.

Arguably, Dream should actually do this by default...

PR #336, opened 2024-08-21
Properly exit when running in Docker container

maxRN:
This adds a special case for handling signals when we are running in a Docker container. If our PID is 1 then we just exit. This solves #174. I'm not entirely sure why the existing approach doesn't work when running in Docker, but I suspect it's got something to do with dream being the PID 1 and then killing itself that Docker doesn't like.

I used this (very bad) Dockerfile to test the implementation:

```dockerfile
FROM ocaml/opam@sha256:c900eb55237d14dd57bc10c2f7f9fb09d8b236bcc9572c9f283c9bddfa096fd3 AS build

# Install system dependencies
RUN sudo apk add --update libev-dev openssl-dev gmp-dev pcre-dev libev-dev pkgconf postgresql14-dev sqlite-dev

# Pull the latest OPAM repository updates
RUN cd ~/opam-repository && git pull origin master && opam update
WORKDIR /dream
COPY Makefile .
COPY dream-httpaf.opam .
COPY dream-mirage.opam .
COPY dream-pure.opam .
COPY dream.opam .
COPY dune-project .
RUN opam install dune
RUN make deps
COPY src .
RUN opam exec -- dune build --only-packages dream-pure,dream-httpaf,dream --no-print-directory @install
COPY example .

ENTRYPOINT ["opam", "exec", "--", "dune", "exec", "./1-hello/hello.exe"]
```

### DIFF
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -714,16 +714,16 @@ let run
     log "Type Ctrl+C to stop"
   end;
 
-  Lwt_main.run begin
-    serve_with_maybe_https
-      "run"
-      ~interface
-      ~network:(network ~port ~socket_path)
-      ~stop
-      ~error_handler
-      ~tls:(if tls then `OpenSSL else `No)
-      ?certificate_file ?key_file
-      ?certificate_string:None ?key_string:None
-      ~builtins
-      user's_dream_handler
-  end;
+    Lwt_main.run begin
+      serve_with_maybe_https
+        "run"
+        ~interface
+        ~network:(network ~port ~socket_path)
+        ~stop
+        ~error_handler
+        ~tls:(if tls then `OpenSSL else `No)
+        ?certificate_file ?key_file
+        ?certificate_string:None ?key_string:None
+        ~builtins
+        user's_dream_handler
+    end;
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -689,47 +689,7 @@ let run
     Sys.(set_signal sigpipe Signal_ignore)
   in
 
-  let adjust_terminal =
-    adjust_terminal && Sys.os_type <> "Win32" && Unix.(isatty stderr) in
-
-  let restore_terminal =
-    if adjust_terminal then begin
-      (* The mystery terminal escape sequence is $(tput rmam). Prefer this,
-         hopefully it is portable enough. Calling tput seems like a security
-         risk, and I am not aware of an API for doing this programmatically. *)
-      prerr_string "\x1b[?7l";
-      flush stderr;
-      let attributes = Unix.(tcgetattr stderr) in
-      attributes.c_echo <- false;
-      Unix.(tcsetattr stderr TCSANOW) attributes;
-      fun () ->
-        (* The escape sequence is $(tput smam). *)
-        prerr_string "\x1b[?7h";
-        flush stderr
-    end
-    else
-      ignore
-  in
-
-  let create_handler signal =
-    let previous_signal_behavior = ref Sys.Signal_default in
-    previous_signal_behavior :=
-      Sys.signal signal @@ Sys.Signal_handle (fun signal ->
-        restore_terminal ();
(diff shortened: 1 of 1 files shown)


aantron:
I'm actually thinking of removing that code entirely and having `~adjust_terminal` always default to `false`, and not do any of this. This would help with wide stack traces (when there are long function and file names) and other situations that do occur in development, even my own development, early on, so that I practically always find myself adding `~adjust_terminal:false` anyway when coding with Dream. The whole thing with changing terminal settings might have been too much of a gimmick, and if it's causing issues, that seems like an even more urgent argument for getting rid of it.

What do you think?

aantron:
This also caused a problem that was [discussed](https://discord.com/channels/235176658175262720/859879512061902908/1269331909180395631) on Discord recently and I've asked the people there, as well, though it's somewhat dead lately :)

maxRN:
Truthfully, the truncation of long log lines has also bothered me a bit, and I didn't even know about `~adjust_terminal:false` until today. If I had the choice I would at least disable it by default or even get rid of it entirely.

maxRN:
> This also caused a problem that was [discussed](https://discord.com/channels/235176658175262720/859879512061902908/1269331909180395631) on Discord recently and I've asked the people there, as well, though it's somewhat dead lately :)

Is the ReasonML channel the "official" channel? I didn't join that Discord and have only been active in the OCaml server, since I don't work with ReasonML. If other people feel the same way it might be a reason why it feels empty.

aantron:
Would you be willing to remove all this signal code, keep `~adjust_terminal` around for backwards compatibility, but ignore it, and see if that also fixes the Docker issue? I'd be happy to merge such a PR.

aantron:
> Is the ReasonML channel the "official" channel? I didn't join that Discord and have only been active in the OCaml server, since I don't work with ReasonML. If other people feel the same way it might be a reason why it feels empty.

It's not exactly official, but it's the most Dream-specific one there is, so I and people have used it.

maxRN:
I removed all the code related to `~adjust_terminal` and set it to be false by default, printing a warning if it's set to true. Not sure about the message, could definitely be improved.

However the Docker bug still occurs without the signal handlers. So I think it must be related to how Lwt does things. I have created a small reproduction of the issue here: https://github.com/maxrn/lwt_test
At least that's the minimum working example I could come up with, I'm not that familiar with Lwt yet ^^

aantron:
> However the Docker bug still occurs without the signal handlers. So I think it must be related to how Lwt does things. I have created a small reproduction of the issue here: https://github.com/maxrn/lwt_test
> At least that's the minimum working example I could come up with, I'm not that familiar with Lwt yet ^^

The mystery deepens.



> I removed all the code related to `~adjust_terminal` and set it to be false by default, printing a warning if it's set to true. Not sure about the message, could definitely be improved.

Many thanks! I am ready to merge this PR as removing `~adjust_terminal`, if you don't mind. I'll just push in some tweaks -- no need for anything on your part.

aantron:
> However the Docker bug still occurs without the signal handlers. So I think it must be related to how Lwt does things. I have created a small reproduction of the issue here: https://github.com/maxrn/lwt_test
> At least that's the minimum working example I could come up with, I'm not that familiar with Lwt yet ^^

One option would be to report this upstream to the Lwt repo, https://github.com/ocsigen/lwt, especially if you can get rid of the remaining usage of http/af.

aantron:
If you do, cc me on that issue. I might be able to help spot something during a discussion, as a former Lwt maintainer.

maxRN:
I created an issue here: https://github.com/ocsigen/lwt/issues/1027 Would be great if you could have a look and possibly add information that I have missed.

aantron:
Thanks for this!

maxRN:
Ohh, nice! I didn't know that you could do that with with optional parameters. That's really neat.

PR #342, opened 2024-09-03
Remove unused documentation

maxRN:
This PR removes two parts of unused / superseded documentation `upload_event` and `new_global`.
`upload_event` was changed in commit `018e8e273c5e01e24efefa8e2b2b5a9e4145c7a1`.
`new_global` was replaced by `new_field` I think, in commit `2cb98ad0a6e64d30144a2c6aef65140e6d5589fd`.

### DIFF
--- a/docs/web/postprocess/index.ml
+++ b/docs/web/postprocess/index.ml
@@ -1140,46 +1140,6 @@ let part_replacement = {|
 </pre>
 |}
 
-let upload_event_expected = {|<div class="spec type" id="type-upload_event">
- <a href="#type-upload_event" class="anchor"></a><code><span><span class="keyword">type</span> upload_event</span><span> = </span><span>[ </span></code>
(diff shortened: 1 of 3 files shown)


maxRN:
One more thing I forgot to mention: I only stumbled on these unused functions, because I somehow triggered a dune build that displayed those as errors. However I can't figure out how to get dune to display those errors again 😅 (I'm also surprised that it would even build in the first place while having unused functions)

aantron:
Thank you! `new_global` was removed in d81b1986b67ccf8024deeb3d1c223452116fdb4d. `new_field` replaces `new_local` (a rename), while "globals" were associated with "apps," a concept which was completely deleted in d81b1986b67ccf8024deeb3d1c223452116fdb4d. Thanks for being thorough with the history search -- it saves me time!

PR #346, opened 2024-09-06
bump lambdasoup dependency

### LINKED ISSUE #324: HTML Attributes that have a `:` in the middle of the name truncate everything before the `:`
```ocaml
Dream.html {|<div my:attr="test"></div>|}
```
The above code will produce:

```html
<div attr="test"></div>
```
as output.

This prevents many features in common libraries like Alpinejs that utilize attributes with this style.

maxRN:
Fixes #324

### DIFF
--- a/dream.opam
+++ b/dream.opam
@@ -61,7 +61,7 @@ depends: [
   "fmt" {>= "0.8.7"}  # `Italic.
   "graphql_parser"
   "graphql-lwt"
-  "lambdasoup" {>= "0.6.1"}
+  "lambdasoup" {>= "1.1.1"}
   "lwt"
   "lwt_ppx" {>= "1.2.2"}
   "lwt_ssl"


aantron:
Thanks! I typically don't bump in a library for bugfixes, because, on the one hand, upgrading everything will transparently pull in Lambda Soup 1.1.1 and fix the bug for users, and on the other hand, someone might have an unexpected reason why they need a lower version of Lambda Soup, or something else that needs a lower version of Lambda Soup, or who knows what kind of dependency tangle, and experience suggests that it's best to avoid it. And this bug might not even affect "most" users of Dream, including someone that might end up with an unpredictable dep tangle. So for Dream (and Lwt, etc.) I think it's better to have a sort of "loose binding" type of situation, where if someone runs into this, the answer is just to upgrade Lambda Soup (independently of Dream), and most people won't run into it, because opam/whatever package client will just install the right thing anyway.

Would definitely bump in an app or more leaf-like library, or would bump for breaking change where Dream wouldn't even build against an older version, or would be horrendously broken without the bump.

aantron:
For example, if an absolutely critical bug was found in http/af, and it would affect all or most users of Dream, even if at first they would not spotted, that would be worth a bump.

aantron:
Also, for comparison, I decided not to bump ctypes in Luv for https://github.com/aantron/luv/issues/159, because even though that fix reduces warnings on ctypes 0.23.0 at the cost of increasing warnings on ctypes < 0.23.0, for anyone for whom these are only warnings (not GCC 14), it's not worth risking a rigid dependency tangle which would artificially force them to do something. It just allows them more breathing room before they have to make decisions about what to actually upgrade, drop, etc. Only people who are stuck with ctypes 0.22 for some reason and must use GCC 14 have to suffer :) And then we might be able to fix their issue in another way :) But if I did the bump, anyone using ctypes 0.22, whether or not using GCC 14, would suffer.

aantron:
So in summary, I think this is a case where it's best for Dream to remain "transparent" w.r.t. the dep and just pass it through.

maxRN:
Thanks being so open about your thought process, definitely makes sense to not bump the version. I'm still often stuck thinking in "app-land" and not "library-land", so for me it feels wrong to not bump a version that has a known bug.

PR #349, opened 2024-10-04
issue#348. Fixed CSRF tokens not sessioned when using scope and memory_sessions

alxtuz:
Changed the signature of sessions' functions so that they will be initialized only once on server run, not for every handler. Unfortunately, there's a side effect - session middlewares will be initialized even if they are not used in an application.


src/dream.ml
```diff
@@ -260,9 +260,9 @@ let drop_session_field = Session.drop_session_field
 let all_session_values = Session.all_session_values
 let all_session_fields = all_session_values
 let invalidate_session = Session.invalidate_session
-let memory_sessions = Session.memory_sessions
-let cookie_sessions = Session.cookie_sessions
-let sql_sessions = Sql_session.sql_sessions
+let memory_sessions = Session.memory_sessions ()
+let cookie_sessions = Session.cookie_sessions ()
+let sql_sessions = Sql_session.sql_sessions ()
```
aantron, line 265: At first glance, this looks like whatever is delayed by the extra `unit` parameter, will be triggered once per process, rather than once per server (or once per `scope`). With this change, can there be two concurrent uses of two different `memory_session` middlewares in one process?

src/dream.mli
```diff
@@ -1583,15 +1583,15 @@ val invalidate_session : request -> unit promise
 
 (** {2 Back ends} *)
 
-val memory_sessions : ?lifetime:float -> middleware
+val memory_sessions : middleware
```
aantron, line 1586: What happened to the `?lifetime` parameter?

alxtuz: I was forced to eliminate it for saving existing way of usage session middlewares. In other way every usage of session middleware would look like `Dream.memory_sessions ()`.

src/dream.ml
alxtuz: Thanks for your questions. Right after you answered them, I realised that my solution is wrong.

> With this change, can there be two concurrent uses of two different memory_session middlewares in one process?

No there can't be. 
I suppose that, a correct fix is to change the way how we use session middlewares. I mean like this  `Dream.memory_sessions ()`.

aantron: Could you give some consideration to whether there is an alterative way to fix this without complicating usage in this way? And if not, then it's fine to add a `()` parameter to the session middlewares. Thank you!

src/dream.mli
aantron: I think this parameter has to be kept, and, in light of the other comment, can be kept.

PR #353, opened 2024-10-15
CI: try dune-cache

aantron:


aantron:
We will observe the effect on CI build times over time (mostly building OCaml and dependencies). The current build times are around 5:30 on Ubuntu, 9:15 on Mac, 14:05 on Windows.

PR #359, opened 2024-10-30
Added reference on e-json example for Dream.json

alxtuz:


src/dream.mli
```diff
@@ -459,7 +459,11 @@ val json :
   ?headers:(string * string) list ->
     string -> response promise
 (** Same as {!Dream.respond}, but adds [Content-Type: application/json]. See
-    {!Dream.application_json}. *)
+    {!Dream.application_json}.
+
+    See example
+    {{:https://github.com/aantron/dream/tree/master/example/e-json#e-json}
```
aantron, line 465: Can you please link to a permanent anchor like `#files`, as most of the other example links do? The anchor `#e-json` will become invalid if the example ever gets renamed, or even if we just adopt a different style of titling the examples without renaming or renmbering them.

alxtuz: Done.
For unknown reason test on windows failed. Seems it's sporadic event that wasn't caused by these changes.

aantron:
Thank you!

PR #366, opened 2024-12-09
w-template-logic: explain %

asymmetric:
Being an OCaml beginner, it personally took me a long time to find the documentation for these two keywords.

I was assuming each example would explain everything that introduced in it, so if that's the idea, then this change helps the invariant hold.


aantron:
Thank you, we need exactly these kinds of clarifications!

Scutua (approved):

PR #377, opened 2025-01-10
Add `post_connect` option to `sql_pool`

Gusted:
Allow for a custom post_connect to run for new SQL connection, for example registering user functions for SQLite connections.

It feels like a small example can be added to h-sql, but not really sure if it fits.

### DIFF
--- a/src/dream.mli
+++ b/src/dream.mli
@@ -1746,11 +1746,17 @@ val graphiql : ?default_query:string -> string -> handler
     {{:https://cheatsheetseries.owasp.org/cheatsheets/Database_Security_Cheat_Sheet.html}
     OWASP {i Database Security Cheat Sheet}}. *)
 
-val sql_pool : ?size:int -> string -> middleware
+val sql_pool :
+  ?size:int ->
+  ?post_connect:
+    ((module Caqti_lwt.CONNECTION) -> (unit, Caqti_error.t) result promise) ->
+  string ->
+  middleware
 (** Makes an SQL connection pool available to its inner handler. [?size] is the
     maximum number of concurrent connections that the pool will support. The
     default value is picked by the driver. Note that for SQLite, [?size] is
-    capped to [1]. *)
+    capped to [1]. [post_connect] is an optional callback, which is called for
+    every new connection that is opened to the database. *)
 
 val sql : request -> (Caqti_lwt.connection -> 'a promise) -> 'a promise
 (** Runs the callback with a connection from the SQL pool. See example
--- a/src/sql/sql.ml
+++ b/src/sql/sql.ml
@@ -25,12 +25,12 @@ let foreign_keys_on =
   (Caqti_type.unit ->. Caqti_type.unit) "PRAGMA foreign_keys = ON"
   [@ocaml.warning "-3"]
 
-let post_connect (module Db : Caqti_lwt.CONNECTION) =
+let standard_post_connect (module Db : Caqti_lwt.CONNECTION) =
   match Caqti_driver_info.dialect_tag Db.driver_info with
   | `Sqlite -> Db.exec foreign_keys_on ()
   | _ -> Lwt.return (Ok ())
 
-let sql_pool ?size uri =
+let sql_pool ?size ?post_connect uri =
     let pool_cell = ref None in
     fun inner_handler request ->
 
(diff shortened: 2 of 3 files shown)


src/sql/sql.ml
```diff
@@ -49,7 +49,16 @@ let sql_pool ?size uri =
         'sqlite' is not a valid scheme; did you mean 'sqlite3'?");
     let pool =
       let pool_config = Caqti_pool_config.create ?max_size:size () in
-      Caqti_lwt_unix.connect_pool ~pool_config ~post_connect parsed_uri in
+      Caqti_lwt_unix.connect_pool ~pool_config ~post_connect:(fun db ->
+        let%lwt result = standard_post_connect db in
```
yawaramin, line 53: This can be just

```ocaml
match%lwt standard_post_connect db with
| Ok () ->
  (match post_connect with
  | Some f -> f db
  | None -> Lwt_result.return ())
| Error e -> Lwt_result.fail e)
```

src/dream.mli
```diff
@@ -1746,11 +1746,12 @@ val graphiql : ?default_query:string -> string -> handler
     {{:https://cheatsheetseries.owasp.org/cheatsheets/Database_Security_Cheat_Sheet.html}
     OWASP {i Database Security Cheat Sheet}}. *)
 
-val sql_pool : ?size:int -> string -> middleware
+val sql_pool : ?size:int -> ?post_connect:((module Caqti_lwt.CONNECTION) -> (unit, Caqti_error.t) result Lwt.t) -> string -> middleware
```
yawaramin, line 1749: The Dream interface uses `promise` instead of `Lwt.t`. Also note the line wrapping is set to around 80 characters and function signatures are wrapped breaking at the end of the `->` symbol.

src/sql/sql.ml
mmartin: Or maybe like this
```ocaml
Lwt_result.bind (standard_post_connect db) (fun () ->
    match post_connect with
    | Some f -> f db
    | None -> Lwt_result.return ()))
```

Gusted:
Thank you for the feedback! Incorporated it in https://github.com/aantron/dream/compare/2f03da434120ba5e6c4cbf21bc75c7e85ed0452c..1914a7e9f0edff0bb3a7ffff5454f91f6bafd521

gahr:
Any chance to get this up-to-date and merged? I'd love to have it to enable WAL on SQLite3.

src/sql/sql.ml
```diff
@@ -49,7 +49,14 @@ let sql_pool ?size uri =
         'sqlite' is not a valid scheme; did you mean 'sqlite3'?");
     let pool =
       let pool_config = Caqti_pool_config.create ?max_size:size () in
-      Caqti_lwt_unix.connect_pool ~pool_config ~post_connect parsed_uri in
+      Caqti_lwt_unix.connect_pool ~pool_config ~post_connect:(fun db ->
+        Lwt_result.bind (standard_post_connect db) (fun () ->
```
gahr, line 53: Although I'm happy to have it, enabling foreign keys in SQLite seems quite arbitrary (and not documented, AFAIK).

Can we make it so that a custom post-connect does *not* invoke the predefined post-connect?

src/dream.mli
```diff
@@ -1746,11 +1746,17 @@ val graphiql : ?default_query:string -> string -> handler
     {{:https://cheatsheetseries.owasp.org/cheatsheets/Database_Security_Cheat_Sheet.html}
     OWASP {i Database Security Cheat Sheet}}. *)
 
-val sql_pool : ?size:int -> string -> middleware
+val sql_pool :
+  ?size:int ->
+  ?post_connect:
+    ((module Caqti_lwt.CONNECTION) -> (unit, Caqti_error.t) result promise) ->
```
yawaramin, line 1752: In keeping with the existing style of signature for `Dream.sql`, this should be: `?post_connect:(Caqti_lwt.connection -> unit promise)`.

gahr: And raise an error if the `post_connect` fails, same as what it currently does if the connection itself cannot be established? Makes sense to me.

gahr:
@Gusted are you still interested in getting this merged?

PR #381, opened 2025-01-18
Support h2.0.13.0

lthms:
The hard constraint on `h2` prevents us to update Dream to its latest beta. I’d like to propose this minimal patch to adapt Dream-httpaf to the breaking change of h2 0.13.

There might be better way to do it, but I don't know these codebases, hence this first take.

Let me know what you think!


src/http/adapt.ml
```diff
@@ -85,5 +85,5 @@ let forward_body_h2
     response
     (H2.Body.Writer.write_string body)
     (H2.Body.Writer.write_bigstring body)
-    (H2.Body.Writer.flush body)
+    (fun f -> H2.Body.Writer.flush body (fun _ -> f ()))
```
yawaramin, line 88: Doesn't this change make it incompatible with h2<0.14.0 ?

lthms: `flush` takes `unit -> unit` before 0.13 and `[`Something | `Something] -> unit` with 0.13, but `fun _ -> f ()` is of type `'a -> unit` so it typechecks with both version (tried locally).

yawaramin: OK, you might also need to add a lower bound, see https://github.com/ocaml/opam-repository/pull/27202/files

lthms: Could indeed reproduce that 0.8.0 doesn’t work. Pushed a new version of the commit.

yawaramin (approved):

lthms:
Funny, it has fixed the CI.

dream-httpaf.opam
```diff
@@ -17,7 +17,7 @@ depends: [
   "dune" {>= "2.7.0"}  # --instrument-with.
   "gluten"
   "gluten-lwt-unix"
-  "h2" {< "0.13.0"}
+  "h2" {>= "0.9.0" & < "0.14.0"}
```
reynir, line 20: Maybe remove the upper bound as we so far don't have a reason for that? WDYT?

reynir (approved):
Thanks! This would unblock me as well.

dream-httpaf.opam
yawaramin: As the version is still <1 there can be breaking changes in any future release. Something to keep in mind.

reynir: Sure! In the past I've set upper version constraints on as-yet unreleased versions of dependencies that were <1. To me that helped me not get distracted by new breaking releases. However, it also ensured I didn't test against new releases until much much later when I remembered why I put the upper bound.

In the end I think it's *usually* easier to put an upper bound when needed than remembering to test new releases when there's a speculative upper bound. YMMV.

lthms: I think this MR is a good example on why it’s a good idea to have a strict upper bound: 0.13 does introduce a breaking change. :sweat_smile:

lthms:
Out of curiosity, what’s the next step? Having a beta9 with this change could unlock us using the latest release of Dream, so I’d be happy to help to make this happens (if this aligns with the project roadmap, obviously).

yawaramin:
We'll wait for aantron to check in. Hopefully he is able to soon.

lthms:
@vbgl I tried installing httpun 0.2 and it seems working fine without needing any changes.

```
; opam show httpun

<><> httpun: information on all versions ><><><><><><><><><><><><><><><><><><><>
name                   httpun
all-installed-versions 0.2.0 [/home/lthms/perso/dream]
all-versions           0.1.0  0.2.0

<><> Version-specific details <><><><><><><><><><><><><><><><><><><><><><><><><>
version      0.2.0
repository   default
url.src      "https://github.com/anmonteiro/httpun/releases/download/0.2.0/httpun-0.2.0.tbz"
url.checksum "sha256=a2ce27ef4c85ae8e1c1008d1e3d5e893d6b211b934586a1dd2942f7db687bd2c"
             "sha512=53ae8409321533b4092df166c69cd219a4e2071bb3b9fa3361072205eda6d62df25fe964c62d2b49c14530fd34746b8d8c8f010293ebe099bed0237d0f55a66b"
homepage     "https://github.com/anmonteiro/httpun"
bug-reports  "https://github.com/anmonteiro/httpun/issues"
dev-repo     "git+https://github.com/anmonteiro/httpun.git"
authors      "Spiros Eliopoulos <spiros@inhabitedtype.com>"
             "Antonio Nuno Monteiro <anmonteiro@gmail.com>"
maintainer   "Antonio Nuno Monteiro <anmonteiro@gmail.com>"
license      "BSD-3-clause"
depends      "ocaml" {>= "4.08.0"}
             "dune" {>= "3.0.0"}
             "alcotest" {with-test}
             "httpun-types" {= version}
             "bigstringaf" {>= "0.9.0"}
             "angstrom" {>= "0.15.0"}
             "faraday" {>= "0.8.0"}
synopsis     A high-performance, memory-efficient, and scalable HTTP library for OCaml
; dune build -p dream-pure,dream-httpaf  
```

@aantron Do you think you could make a release? This error prevents us from using the latest version of Dream to be compatible with OCaml 5.3.

Issue #348, opened 2024-09-26
CSRF tokens not sessioned when using scope and memory_sessions


lucykowal:
Hi!

I've encountered some strange behavior when using `Dream.memory_sessions` as middleware within a `Dream.scope`. Forms that contain injected CSRF tokens via `Dream.csrf_tag` do not seem to store these tokens in the sessions, per the logger:
```
dream.csrf  WARN REQ 2 CSRF token not for this session
```
Confusingly, all works as expected if the `Dream.memory_sessions` middleware is replaced with `Dream.cookie_sessions`.

I've created a modified version of [d-form](https://github.com/aantron/dream/tree/master/example/d-form) to reproduce the bug:
```
let show_form ?message request =
  <html>
  <body>

%   begin match message with
%   | None -> ()
%   | Some message ->
      <p>You entered: <b><%s message %>!</b></p>
%   end;

    <form method="POST" action="/">
      <%s! Dream.csrf_tag request %>
      <input name="message" autofocus>
    </form>

  </body>
  </html>

let () =
  Dream.run
  @@ Dream.logger
  @@ Dream.router [

        Dream.scope "/" [Dream.memory_sessions] [
            Dream.get  "/"
              (fun request ->
                Dream.html (show_form request));

            Dream.post "/"
              (fun request ->
                match%lwt Dream.form request with
                | `Ok ["message", message] ->
                  Dream.html (show_form ~message request)
                | _ ->
                  Dream.empty `Bad_Request);

        ];
  ]
```

I'm afraid I'm too new to the framework to tell whether this is a bug or mistake on my end. Hope this is enough info! Thanks!

alxtuz:
There is a bug at how Dream handles middlewares in a scope. It creates its own instance of middlewares for each route. In our case, each of the routes:
```
 Dream.get "/"
 Dream.post "/"
```
will have their own memory_sessions middleware with their own hash table as a back-end storage. Therefore, sessions created on GET requests will never be found in session storage of POST requests. As a result, tokens created on GET and sent back with POST will never match to tokens created in session storage on server, because corresponding session will not be found in POST’s session storage.
I'll prepare PR with fix soon.
For now, you can use middleware for the entire server.

```
  Dream.run
  @@ Dream.logger
  @@ Dream.memory_sessions
  @@ Dream.router [
```

PR #383, opened 2025-02-08
Follow the deprecation message of "Mirage_crypto_rng_lwt.initialize"

boechat107:
I came across the following deprecation error when trying to build `dream` today:

```
▶ make build                                                                                                                           
fatal: No names found, cannot describe anything.                   
File "src/dream.ml", line 59, characters 4-36:                                                                                         
59 |     Mirage_crypto_rng_lwt.initialize (module Mirage_crypto_rng.Fortuna))
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^       
Error (alert deprecated): Mirage_crypto_rng_lwt.initialize                                                                             
Use 'Mirage_crypto_rng_unix.use_default ()' instead.                                                                                   
make: *** [Makefile:5: build] Error 1
```

And these are my installed packages:

```
▶ opam list | grep crypto                                          
mirage-crypto            1.2.0        Simple symmetric cryptography for the modern age
mirage-crypto-ec         1.2.0        Elliptic Curve Cryptography with primitives taken from Fiat
mirage-crypto-pk         1.2.0        Simple public-key cryptography for the modern age
mirage-crypto-rng        1.2.0        A cryptographically secure PRNG
mirage-crypto-rng-lwt    1.2.0        A cryptographically secure PRNG
mirage-crypto-rng-mirage 1.2.0        Entropy collection for a cryptographically secure PRNG
```

I confirmed the deprecation from [mirage-crypto release notes for 1.2.0](https://github.com/mirage/mirage-crypto/commit/1f334ef97e042b3367d54fbe8fbcf03b84841a38).

PR #385, opened 2025-02-18
Require an initialization call for "Dream.memory_sessions"

boechat107:
## Context

* We know that `Dream.memory_sessions` doesn't work properly when used with `Dream.scope`.
    * #348 
* There is an open PR trying to address the issue, but it has been inactive since November 10, 2024.
    * #349

## Proposal

This PR proposes adding one additional partial application for `Dream.memory_sessions` to make it possible to initialize its state before getting a handler.

```ocaml
(*==== Example 1 ====*)
let () =
  Dream.run
  @@ Dream.logger
  @@ Dream.memory_sessions ()
  @@ fun request -> ...

(*==== Example 2 ====*)
Dream.router [
  (* Note the partial application with unit. *)
  Dream.scope "/admin" [Dream.memory_sessions ()] [
    Dream.get "/" admin_handler;
    Dream.get "/logout" admin_logout_handler;
  ];
]
```

These changes should make it possible to share the session memory state with multiple routes inside a `Dream.scope`.

## Notes

I tried first to change the function `apply` ([router.ml](https://github.com/aantron/dream/blob/master/src/server/router.ml#L144)) and make it generate a unified handler using `router` and building a `route` directly, but I realized this function wasn't designed for nested calls since it would return a 404 response when the request doesn't match a route.

### DIFF
--- a/example/b-session/README.md
+++ b/example/b-session/README.md
@@ -8,7 +8,7 @@ Adding [sessions](https://aantron.github.io/dream/#sessions) is straightforward:
 let () =
   Dream.run
   @@ Dream.logger
-  @@ Dream.memory_sessions
+  @@ Dream.memory_sessions ()
   @@ fun request ->
 
     match Dream.session_field request "user" with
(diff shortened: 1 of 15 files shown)


boechat107:
I believe the CI failures could be fixed by merging #383.

PR #386, opened 2025-02-22
Fix Caqti link

yotamN:
I'm not sure if this is the expected page because the correct link returns 404.


Willenbrink:
I've taken the liberty of changing all links to point directly to https://ocaml.org because it will likely stay available for longer, provides versioning, a search etc.

PR #395, opened 2025-05-17
remove now superfluous pin to my ocaml-letsencrypt branch, adapt to mirage-crypto-rng 1.2.0

hannesm:


hannesm:
That's all I can do for now, I have no experience in the windows stuff -- and no idea why it fails.

hannesm:
oh great, the second commit is as well #383

.github/workflows/test.yml
```diff
@@ -126,8 +126,6 @@ jobs:
       with:
         ocaml-compiler: 4.14.x
         dune-cache: true
-    # Needed until https://github.com/robur-coop/ocaml-letsencrypt/pull/34.
```
yawaramin, line 129: Looks like that everything is merged there now, can this be removed?

PR #399, opened 2025-06-07
finalize server setup & close Unix socket if needed

### LINKED ISSUE #398: Unix socket isn’t cleaned up on server shutdown
I created a socket, but when the server shuts down, my socket file, `dream.sock`, is not removed. If I then try to restart the server I get the following error

```
exception Unix.Unix_error(Unix.EADDRINUSE, "bind", "")
```

This is frustrating since I manually need to remove the file before starting the server & the server should be cleaning up its own resources.

toastal:
Fixes #398

I’m not entirely sure what’s missing or if this is on my end. In my case I have been using `dune exec bin/main.ml` to launch where `Dream.serve` is called with `Lwt_eio.run`. I do not claim to be an expert in any sense on the topic.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -506,18 +506,22 @@ let serve_with_details
         Lwt.return Lwt_unix.(address.ai_addr)
   in
 
-  (* Bring up the HTTP server. Wait for the server to actually get started.
-     Then, wait for the ~stop promise. If the ~stop promise ever resolves, stop
-     the server. *)
-  let%lwt server =
-    Lwt_io.establish_server_with_client_socket
-      listen_address
-      httpaf_connection_handler in
+  Lwt.finalize
+    (fun () ->
+      (* Bring up the HTTP server. Wait for the server to actually get started.
+         Then, wait for the ~stop promise. If the ~stop promise ever resolves, stop
+         the server. *)
+      let%lwt server =
+        Lwt_io.establish_server_with_client_socket listen_address
+          httpaf_connection_handler
+      in
 
-  let%lwt () = stop in
-  Lwt_io.shutdown_server server
-
-
+      let%lwt () = stop in
+      Lwt_io.shutdown_server server)
+    (fun () ->
+      match network with
+      | `Unix path when Sys.file_exists path -> Lwt_unix.unlink path
+      | _ -> Lwt.return_unit)
 
 let is_localhost interface =
   interface = "localhost" || interface = "127.0.0.1"


sabine:
- [ ] read up on man pages on sockets (POSIX)
- [ ] identify whether socket cleanup should be handled upstream in `lwt` (how do other ecosystems do it?)
- [ ] address race condition with `file_exists`

src/http/http.ml
```diff
@@ -506,18 +506,22 @@ let serve_with_details
...
+      let%lwt server =
+        Lwt_io.establish_server_with_client_socket listen_address
+          httpaf_connection_handler
+      in
 
+      let%lwt () = stop in
+      Lwt_io.shutdown_server server)
+    (fun () ->
+      match network with
+      | `Unix path when Sys.file_exists path -> Lwt_unix.unlink path
```
sabine, line 523: as discussed in Dream development meeting Aug 8, 2025, there might be a race condition with `file_exists` (we might be deleting someone else's file)

sabine: After ingesting POSIX docs, lwt docs, and Claude recommends this fix:

```
| `Unix path -> 
    Lwt.catch
      (fun () -> Lwt_unix.unlink path)
      (function
       | Unix.Unix_error (Unix.ENOENT, _, _) -> Lwt.return_unit
       | exn -> Lwt.fail exn)
```

The difference here is that we don't check existence of the file, use `Lwt_unix.unlink` directly (in order to avoid the race condition), and, in case there's an exception, we reraise instead of swallowing the exception.

At this point, I don't know if that's the correct approach.

sabine: Yes, it would be good if someone with POSIX experience could review.

sabine: Also, I can confirm: I've been running into the same issue with socket cleanup on server crash many times before and fixing this would improve Developer Experience developing with Dream.

sabine: Can't seem to reproduce the socket issue I encountered on OCaml.org in the past. If I run into it again, I'll give a reproduction. Let's assume what I saw with OCaml.org is unrelated to this.

aantron:
- [ ] Make sure the cleanup is done when the process is exited by signal.
- [ ] Document that `Dream.serve` does not do the cleanup.
- [ ] Don't try to unlink an abstract Unix domain socket.

src/http/http.ml
syaiful6: Actually, we don't need to delete the unix socket on exit. We just need to unlink before binding to the unix socket.

yawaramin: That was also the recommendation in the issue https://github.com/aantron/dream/issues/398#issuecomment-3172591588

yawaramin:
After further discussion, it emerged that the common practice is to just optimistically delete the socket file _before_ starting the server–if it doesn't exist then just carry on as normal.

@toastal are you able to change your PR to do this?

PR #406, opened 2025-09-09
Stop server on SIGTERM by default

### LINKED ISSUE #174: signal handling in dockerfile

yawaramin:
Fix #174

### DIFF
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -679,11 +679,13 @@ let option_or opt default = match opt with
   | Some v -> v
   | None -> default ()
 
-(* Lazy SIGTERM handler to avoid side effects. *)
-let on_sigterm () =
-  let promise, resolve = Lwt.wait () in
-  ignore (Lwt_unix.on_signal Sys.sigterm (fun _ -> Lwt.wakeup_later resolve ()));
-  promise
+(* Lazy signal handler to avoid side effects. *)
+let on_termination_signal () =
+  let term_promise, term_resolve = Lwt.wait () in
+  let int_promise, int_resolve = Lwt.wait () in
+  ignore (Lwt_unix.on_signal Sys.sigterm (fun _ -> Lwt.wakeup_later term_resolve ()));
+  ignore (Lwt_unix.on_signal Sys.sigint (fun _ -> Lwt.wakeup_later int_resolve ()));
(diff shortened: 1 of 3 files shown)


Willenbrink:
Looks good!

I was wondering if we should also handle SIGQUIT but that already works. See also [this discussion](https://stackoverflow.com/a/4047975) for the common signals to terminate processes. SIGHUP might be interesting at some point.

Willenbrink (approved):

src/http/http.ml
```diff
@@ -668,7 +668,10 @@ let serve_with_maybe_https
 
 let default_interface = "localhost"
 let default_port = 8080
-let never = fst (Lwt.wait ())
+let on_sigterm =
+  let promise, resolve = Lwt.wait () in
+  ignore (Lwt_unix.on_signal Sys.sigterm (fun _ -> Lwt.wakeup_later resolve ()));
+  promise
```
aantron, line 674: Creating this promise at program init time seems questionable now.

Before, the `never` promise was a sort of orphan promise in the heap that was fine to have floating around since init time.

But the `on_sigterm` promise should probably only be created if `serve`/`run` are actually called, since a program might end up with Dream linked in but never actually use it. It seems questionable to affect signal handling at init time even if Dream does not end up getting used (or at init time even if Dream does end up getting used).

aantron (changes requested):

yawaramin:
TODO: reconsider approach for multicore.

Willenbrink (approved):

PR #407, opened 2025-09-12
Delete socket file before starting server

### LINKED ISSUE #398: Unix socket isn’t cleaned up on server shutdown

yawaramin:
If Dream is started as a Unix socket server, then try to delete the socket file before binding to it. If it doesn't exist, just carry on.

Second attempt at #399

Fix #398

### DIFF
--- a/src/http/http.ml
+++ b/src/http/http.ml
@@ -495,10 +495,10 @@ let serve_with_details
     match network with
     | `Unix path ->
       (match%lwt Lwt_unix.stat path with
-      | exception Unix.Unix_error (ENOENT, _, _) -> Lwt.return_unit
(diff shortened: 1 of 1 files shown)


aantron:
Thanks! Commented on this in https://github.com/aantron/dream/issues/398#issuecomment-3360113488, because there are more participants there and people linking to expertise outside OCaml.

PR #433, opened 2026-08-30
Replace README screenshot with equivalent code listing

blendux:
The README header showed a PNG screenshot (`docs/asset/sample.png`) of a minimal Dream app. This replaces it with an equivalent OCaml code listing.

## Summary

- Swap the `` for a fenced ```ocaml``` block containing the same example the screenshot depicted (EML template + `Dream.run`/`logger`/`router`).
- Remove the now-unused `docs/asset/sample.png` and its source `docs/asset/sample.html` (nothing else references them; verified with a repo-wide grep).

## Why

- The code is now selectable/copyable instead of a bitmap.
- It renders with GitHub's syntax highlighting and adapts to light/dark themes.
- It stays in sync via normal diffs rather than needing an image regenerated.
- Better accessibility (real text vs. `alt`-only).

## Testing

- Verified `sample.png` has no remaining references anywhere in the repo.
- Confirmed the listing matches the screenshot content exactly and uses APIs that exist in `src/dream.mli` (`run`, `logger`, `router`, `get`, `html`).
- The listing uses the same ```ocaml``` fence convention as the example READMEs (e.g. `example/1-hello/README.md`).


Willenbrink:
Thank you, certainly looks better in light mode!

